PR #10, opened 2018-11-23
Allocation all code (do not merge - reviewed parts are in "allocation" branch).

ppalucki:
Long-lived branch that includes allocation feature:

TODO:
- [x] refactor runners to extract common functinoallity
- [x] RDT resources allocation
- [x] RDT independent monitoring and allocations (MON_GROUPS cs CTRL_GROUPS)
- [x] cgroup based resources allocation
- [x] extra profiling metrics
- [x] refactor labeling (as metrics package)
- [x] calculating resulting allocations
- [x] encoding allocations as metrics
- [x] unit tests:
  - [x] Cgroup object
  - [x] ResGroup object
  - [x] resulting allocations 
  - [x] AllocationRunner/ContainerManager and Containers
- [x] documentation (another PR)
- [x] simple allocator example
- [x] extend platform information 
- [x] validation of returned allocations (RDT )
- [x] validation of returned allocations (non RDT (shares, quota))
- [x] default RDT allocation and proper regroup cleaning in the initialization phase
- [x] implemented abstract based classed based serialization and merging of allocation value
- [x] functional/e2e tests on dedicated hardware
   - [x] removing and add workloads in the same ResGroups
   - [x] garbage collecting of unused ResGroups

Bugs:
- [x] race condition for invalid pids, while synchronizing cgroups/tasks with resgroup/tasks
- [x] reappearing root/mon_groups


owca/allocators.py
```diff
@@ -48,13 +90,16 @@ class RDTAllocation:
 class AllocationConfiguration:
 
     # Default value for cpu.cpu_period [ms] (used as denominator).
-    cpu_quota_period: int = 100
+    cpu_quota_period: int = 1000
```
ss7pro: Why we're changing this?

configs/mesos_example_allocator.yaml
```diff
@@ -0,0 +1,31 @@
+# Copyright (c) 2018 Intel Corporation
```
bwplotka: Hm.. Year looks wrong to me (:

owca/allocators.py
ppalucki: because it is the minmal value

configs/mesos_example_allocator.yaml
ppalucki: It was committed to finish, last year :) I don't want to create proof that we are out of schedule ! :innocent:

felidadae:
The branch for the PR is used by PRC team and is frozen.

ppalucki:
super seeded #10

Issue #25, opened 2019-01-09
owca crashes while restarting the mesos tasks


WangJialei-A:
```
Traceback (most recent call last):
  File ".bootstrap/pex/pex.py", line 352, in execute
  File ".bootstrap/pex/pex.py", line 284, in _wrap_coverage
  File ".bootstrap/pex/pex.py", line 315, in _wrap_profiling
  File ".bootstrap/pex/pex.py", line 397, in _execute
  File ".bootstrap/pex/pex.py", line 495, in execute_entry
  File ".bootstrap/pex/pex.py", line 510, in execute_pkg_resources
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/main.py", line 115, in main
    runner.run()
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/logger.py", line 112, in __trace
    rv = func(*args, **kw)
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/runner.py", line 188, in run
    task_measurements = container.get_measurements()
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/containers.py", line 56, in get_measurements
    self.cgroup.get_measurements(),
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/cgroups.py", line 39, in get_measurements
    with open(os.path.join(self.cgroup_fullpath, CPU_USAGE)) as \
[...]

PR #27, opened 2019-01-10
Fix race condition: the mesos task killed during runner iteration loop.

felidadae:
Fixing https://github.com/intel/owca/issues/25.
Tested on a real cluster, by stopping run by the debugger, killing mesos jobs and rerunning the application.

### DIFF
--- a/owca/resctrl.py
+++ b/owca/resctrl.py
@@ -183,4 +183,3 @@ class ResGroup:
             os.rmdir(self.resgroup_dir)
         except FileNotFoundError:
             log.debug('cleanup: directory already does not exist %s', self.resgroup_dir)
-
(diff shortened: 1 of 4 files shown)


owca/resctrl.py
```diff
@@ -178,5 +178,9 @@ def get_measurements(self) -> Measurements:
...
 
     def cleanup(self):
-        log.log(logger.TRACE, 'resctrl: rmdir(%s)', self.resgroup_dir)
-        os.rmdir(self.resgroup_dir)
+        try:
+            log.log(logger.TRACE, 'cleanup: rmdir(%s)', self.resgroup_dir)
+            os.rmdir(self.resgroup_dir)
+        except FileNotFoundError:
+            log.debug('cleanup: directory already does not exist %s', self.resgroup_dir)
+
```
iwankgb:
```suggestion
```

Issue #24, opened 2019-01-09
owca crashes while parsing the result from mesos.


WangJialei-A:
```
Traceback (most recent call last):
  File ".bootstrap/pex/pex.py", line 352, in execute
  File ".bootstrap/pex/pex.py", line 284, in _wrap_coverage
  File ".bootstrap/pex/pex.py", line 315, in _wrap_profiling
  File ".bootstrap/pex/pex.py", line 397, in _execute
  File ".bootstrap/pex/pex.py", line 495, in execute_entry
  File ".bootstrap/pex/pex.py", line 510, in execute_pkg_resources
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/main.py", line 115, in main
    runner.run()
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/logger.py", line 112, in __trace
    rv = func(*args, **kw)
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/runner.py", line 160, in run
    tasks = self.node.get_tasks()
  File "/root/.pex/install/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl.772bfaf4d1d8d5ea3921f399d71390ea5d1bb83f/owca-0.1.dev213+g69633f1.d20190107-py3-none-any.whl/owca/mesos.py", line 108, in get_tasks
    statuses = launched_task['statuses']
KeyError: 'statuses'
```

felidadae:
We will today add fix for this problem. We have already investigated it.

felidadae:
PR fixing: https://github.com/intel/owca/pull/28

PR #28, opened 2019-01-11
Fix bug: owca crash while reading data from mesos.

felidadae:
Fixing https://github.com/intel/owca/issues/24

### DIFF
--- a/owca/mesos.py
+++ b/owca/mesos.py
@@ -105,10 +105,11 @@ class MesosNode(Node):
             return []
 
         for launched_task in state['get_state']['get_tasks']['launched_tasks']:
+            if launched_task['state'] != MESOS_TASK_STATE_RUNNING:
+                continue
+
             statuses = launched_task['statuses']
             last_status = statuses[-1]  # Assume the last on is the latest state # TODO: confirm
-            if last_status['state'] != MESOS_TASK_STATE_RUNNING:
-                continue
 
             if 'executor_pid' not in last_status['container_status']:
                 log.warning("'executor_pid' not found in container status for task %s on agent %s",


owca/mesos.py
```diff
@@ -105,10 +105,11 @@ def get_tasks(self):
             return []
 
         for launched_task in state['get_state']['get_tasks']['launched_tasks']:
+            if launched_task['state'] != MESOS_TASK_STATE_RUNNING:
+                continue
+
             statuses = launched_task['statuses']
             last_status = statuses[-1]  # Assume the last on is the latest state # TODO: confirm
-            if last_status['state'] != MESOS_TASK_STATE_RUNNING:
```
iwankgb: Do not remove this check, please.

tests/test_mesos.py
```diff
@@ -61,3 +61,26 @@ def test_missing_executor_pid(post):
     tasks = node.get_tasks()
 
     assert len(tasks) == 0
+
+
+def test_missing_statuses():
+    mesos_json_response = json.load(
+        open(relative_module_path(__file__,
+                                  'fixtures/missing_executor_pid_in_mesos_response.json')))
```
iwankgb: I believe that we should prepare separate file for each failure scenario. It would make it easier to understand. Just fetch output from Mesos agent and modify to meet our needs. 
Consider leaving just one container there too (to make it easier to digest for a human being).

iwankgb: Having separate file for each case will allow us to use `@pytest.mark.parametrize` with multiple values and reuse same test body.

felidadae: I am not sure if that is good solution as we end up with many files, where taking into consideration their size its not simply to say what are the differences. Whereas having one file with mesos response and adjusting it for test case, the adjusting procedure documents the differences. However I made suggested changes in the code.

PR #35, opened 2019-01-18
Trace decorator improvements

ppalucki:
``@trace`` decorator is not always verbose anymore and allows more granular control on what
and when is displayed.

When applied on function, it will be default:
- show arguments and call entry and exit messages with time taken when TRACE level is enabled with TRACE level
- shows call entry and exist message with time taken when DEBUG level is enabled with DEBUG level

Forcing or disabling show_arguments can be specified by verbose argument.
If vebose is set to False, logs will never include arguments (to verbose)
If verbose is set to True, logs will always include arguments (assuming it is appropriate for DEBUG level)


owca/logger.py
```diff
@@ -100,17 +110,39 @@ def some_function(x):
...
+    [TRACE] owca.some_module: <- some_function() (1.5s)
 
     """
     def _trace(func):
         def __trace(*args, **kw):
             s = time.time()
-            log.log(TRACE, '-> %s(args=%r, kw=%r)', func.__name__, args, kw)
+
+            if verbose is not None:
+                show_args = verbose
```
iwankgb: s/show_args/log_input_output/

PR #38, opened 2019-02-04
[kubernetes:1] minimal DetectRunner scenario

felidadae:
Branch kubernetes-support is based on 9d78321ab7ae5f88b61dbed6a65830c426c09e98 from ppalucki/allocation-wip. I will each few days merge master and allocations branch into this branch to stay up to date.

This branch contains initial support for kubernetes: only DetectRunner is supported and only one container per pod is supported.

Please notice that commits history is ordered and commits are well seperated to improve readibility.

Please provide only high level comments - kubernetes-support is not a production branch and probably will become long-living branch.


ppalucki:
Just one comments, (non blocking) :

We should consider allowing node scoped pods discovery based on undocumented kubelet API ([here](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/server/server.go#L283) e.g. http://localhost:10255/pods 
Then for local discovery,  I would like to not depend on official kubernetes client (you cannot) and assume that local connection is secure and authorized.

We can always add that later as another option.

felidadae:
@TODO some of the commits should be cherry picked to master.

PR #41, opened 2019-02-06
allocations [1]: allocators and base allocations modules

ppalucki:


owca/allocations.py
```diff
@@ -0,0 +1,197 @@
+# Copyright (c) 2018 Intel Corporation
```
iwankgb:
```suggestion
# Copyright (c) 2019 Intel Corporation
```

tests/allocations/test_allocations_dict.py
```diff
@@ -0,0 +1,103 @@
+# Copyright (c) 2018 Intel Corporation
```
iwankgb:
```suggestion
# Copyright (c) 2019 Intel Corporation
```

tests/allocations/test_allocations_numeric.py
```diff
@@ -0,0 +1,76 @@
+# Copyright (c) 2018 Intel Corporation
```
iwankgb:
```suggestion
# Copyright (c) 2019 Intel Corporation
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+        self.common_labels = common_labels
+
+    def update_labels(self, metrics):
+        """Update labels values inplace."""
+        for metric in metrics:
+            metric.labels.update(**self.common_labels)
+
+
+class BoxedNumeric(AllocationValue):
+    """ AllocationValue for numeric like values (floats and ints).
```
iwankgb:
```suggestion
    """ AllocationValue for numeric values (floats and ints).
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+    Wrapper for floats and integers.
+    If min_value is None then it becomes negative infinity (default is 0).
+    If max_value is None then it becomes infinity (default is None(infinity).
+    """
+    # Defines default how sensitive in terms of
```
iwankgb:
```suggestion
    # Defines precision of number comparison. See: math.isclose()
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+    # float precision are changes are detected.
```
iwankgb:
```suggestion
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+    FLOAT_VALUES_CHANGE_DETECTION = 0.05
+
+    def __init__(self, value: Union[float, int],
+                 common_labels: Dict[str, str] = None,
+                 min_value: Optional[Union[int, float]] = 0,
+                 max_value: Optional[Union[int, float]] = None,
+                 float_value_change_sensitivity=FLOAT_VALUES_CHANGE_DETECTION,
```
iwankgb:
```suggestion
                 value_change_sensitivity: float = VALUE_CHANGE_SENSITIVITY,
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+
+class BoxedNumeric(AllocationValue):
+    """ AllocationValue for numeric like values (floats and ints).
+    Wrapper for floats and integers.
+    If min_value is None then it becomes negative infinity (default is 0).
+    If max_value is None then it becomes infinity (default is None(infinity).
+    """
+    # Defines default how sensitive in terms of
+    # float precision are changes are detected.
+    FLOAT_VALUES_CHANGE_DETECTION = 0.05
```
iwankgb:
```suggestion
    VALUE_CHANGE_SENSITIVITY = 0.05
```
(you will use it to compare integers too).

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+
+    def __init__(self, value: Union[float, int],
+                 common_labels: Dict[str, str] = None,
+                 min_value: Optional[Union[int, float]] = 0,
+                 max_value: Optional[Union[int, float]] = None,
+                 float_value_change_sensitivity=FLOAT_VALUES_CHANGE_DETECTION,
+                 ):
+        assert isinstance(value, (float, int))
+        self.value = value
+        self.float_value_change_sensitivity = float_value_change_sensitivity
```
iwankgb: s/float_value_change_sensitivity/value_change_sensitivity/

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+
+            value_changed = (self != current)
+
+        if value_changed:
+            # For floats merge is simple, is value is change, the
+            # new_value just become target and changeset
+            # target and changeset (overwrite policy)
+            return self, self
+        else:
+            # If value is not changed, then is assumed current value is the same as
```
iwankgb:
```suggestion
            # If value is not changed, then value is the same as
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            # There is no old value, so there is a change
+            value_changed = True
+        else:
+            # If we have old value compare them.
+            assert isinstance(current, BoxedNumeric)
+
+            value_changed = (self != current)
+
+        if value_changed:
+            # For floats merge is simple, is value is change, the
```
iwankgb:
```suggestion
            # If value is changed then self becomes target state
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            # new_value just become target and changeset
```
iwankgb:
```suggestion
            # and changeset.
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            # target and changeset (overwrite policy)
```
iwankgb:
```suggestion
```

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+    def __init__(self, common_labels):
+        self.common_labels = common_labels
+
+    def update_labels(self, metrics):
+        """Update labels values inplace."""
+        for metric in metrics:
+            metric.labels.update(**self.common_labels)
+
+
+class BoxedNumeric(AllocationValue):
```
iwankgb: This class must be an abstract one (as perform_allocations() is not implemented).

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            # For floats merge is simple, is value is change, the
+            # new_value just become target and changeset
+            # target and changeset (overwrite policy)
+            return self, self
+        else:
+            # If value is not changed, then is assumed current value is the same as
+            # new so we can return any of them (lets return the new one) as target
+            return current, None
+
+    def perform_allocations(self):
```
iwankgb: Can we drop this function completely when class is marked as abstract?

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            value.validate()
+
+
+class LabelsUpdater:
+    """Helper object to update metrics."""
+
+    def __init__(self, common_labels):
+        self.common_labels = common_labels
+
+    def update_labels(self, metrics):
```
iwankgb: Can you test this method, please?

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+            current_value = current.get(key)
+
+            if current_value is None:
+                # There is no current value, new is used as both target and changeset.
+                target[key] = new_value
+                changeset[key] = new_value
+            else:
+                # Both exists - recurse into values.
+                assert isinstance(current_value, AllocationValue)
+                target_value, value_changeset = new_value.calculate_changeset(current_value)
```
iwankgb: Haven't we assumed that we do not need recursion any more? As far as I understand there are only values inside `AllocationsDict` rather than other instances of `AllocationsDict`.

owca/allocations.py
```diff
@@ -0,0 +1,197 @@
...
+                assert isinstance(target_value, AllocationValue)
+                assert isinstance(value_changeset, (type(None), AllocationValue))
+                target[key] = target_value
+                if value_changeset is not None:
+                    changeset[key] = value_changeset
+
+        # If there is no fields in changeset dict return None
```
iwankgb:
```suggestion
        # If there are no fields in changeset dict return None
```

owca/allocators.py
```diff
@@ -11,26 +11,31 @@
...
+log = logging.getLogger(__name__)
 
-class AllocationType(str, Enum):
 
+class AllocationType(str, Enum):
     QUOTA = 'cpu_quota'
     SHARES = 'cpu_shares'
     RDT = 'rdt'
 
+    def __repr__(self):
```
iwankgb: What is this for?

tests/allocations/test_allocations_dict.py
```diff
@@ -0,0 +1,103 @@
...
+         dict(t1={'a': 2.1}), dict(t1={'a': 2.1})),
+        (dict(t1={'a': 2}), dict(t1={'a': 2}),
+         dict(t1={'a': 2}), None),
+        (dict(t1={'a': 1}), dict(t1={'b': 2}, t2={'b': 3}),
+         dict(t1={'a': 1, 'b': 2}, t2={'b': 3}), dict(t1={'b': 2}, t2={'b': 3})),
+    ]
+)
+def test_allocations_dict_merging(current, new,
+                                  expected_target, expected_changeset):
+    def convert_to_allocations_dict(d: dict):
```
iwankgb: I don't think it needs to be a closure. I think it should be promoted to independent private function.

tests/allocations/test_allocations_numeric.py
```diff
@@ -0,0 +1,76 @@
...
+            (2.0, None, 3.0, 0.00001, None),
+            (2.0, 1, None, 0.00001, None),
+    )
+)
+def test_boxed_numeric_validation(value, min_value, max_value, float_value_change_sensitivity,
+                                  expected_error):
+    boxed_value = BoxedNumeric(value, min_value=min_value,
+                               max_value=max_value,
+                               float_value_change_sensitivity=float_value_change_sensitivity)
+    if expected_error:
```
iwankgb: Instead of adding a conditional statement to test code I would prefer separate test case for validation failure scenarios.

tests/allocations/test_allocations_dict.py
ppalucki: closure by definition is when one is using local context - this function isn't - 
it is not a closure just a helper function inside a test for only this test - it has no purpose outside this test scope 

if you insists I can move it out, but I really wanted test cases to have not other dependencies that owca/testings common helpers or real code they use

owca/allocations.py
ppalucki: We haven't removed recursion completely - still calculating changeset over nested dicts assumes that object one-level deeper is AllocationValue (the same is for generate_metrics, validation and so on).

But we don't use recursion explictly because we derive from AllocationDict to create TasksAllocationsValues (first level) and TasksAllocationValues(second level). 
It is more verbose but more explicit at the same time.

Recursion that was hard to comprehend, were happening when previously implict conversion was happening in constructor and we could expect very long leaves (like 'AD' -> 'AD' -> 'ErrorContext' -> 'CommonLabels' -> 'InvalidArgument' -> 'RDTAllocation')

Now it is always two levels deep, but still use the idea that calculating changeset, validation, metrics as exactly the same function on both levels (but bound/derived to different specialized classes).

owca/allocations.py
ppalucki: replace with no-op implmentation - I don't want it abstract because is need for calculating changes, metrics and performing tests without need to implement subclass doing nothing.

If you insist I can make it abstract and the need to add BoxNumericDummy to all tests.

owca/allocations.py
ppalucki: done

owca/allocations.py
ppalucki: replaced with no-op function (see comment above)

owca/allocators.py
ppalucki: when logs are enabled at debug/trace level
all structures will be nicely printed like this:
```python
'taskid1': {
  'cpu_shares': 32,
  'rdt': RDTAllocation(name='asdf', l3),
}}
```

instead of:

```python
'taskid1': {
  <AllocationValue.CPU_SHARES> : 32,
  <AllocationValue.RDT>: RDTAllocation(name='asdf', l3),
}}
```

tests/allocations/test_allocations_numeric.py
ppalucki: as you wish - done

owca/allocations.py
ppalucki: done test_labels_updater in test_allocations_dict.py

owca/allocations.py
iwankgb: I think it would be better approach as it would prevent developers from attempting to use it directly in future. Use `BoxNumericDummy` in tests, please.

ppalucki: done

PR #44, opened 2019-02-12
allocations [3]: resctrl and cgroups implementantion

ppalucki:


ppalucki:
too big superseded by #46

PR #46, opened 2019-02-12
allocations [3]: cgroups allocations

ppalucki:


owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+
+    def set_normalized_shares(self, shares_normalized):
+        """Store shares normalized values in cgroup files system. For denormalization
+        we use reverse formule to _get_normalized_shares."""
+        assert self.allocation_configuration is not None, \
+            'allocation configuration cannot be used without configuration!'
+
+        shares = int(shares_normalized * self.allocation_configuration.cpu_shares_unit)
+        if shares < 2:
+            shares = 2
```
pskrzyns: Can u use constant instead of 2 ?
Also dont you think that ```shares = max([int(shares_normalized * self.allocation_configuration.cpu_shares_unit), 2])``` looks better ?

owca/cgroup_allocations.py
```diff
@@ -0,0 +1,53 @@
...
+    def __init__(self, normalized_shares: float, container: Container, common_labels: dict):
+        self.normalized_shares = normalized_shares
+        self.cgroup = container.cgroup
+        super().__init__(value=normalized_shares, common_labels=common_labels, min_value=0)
+
+    def generate_metrics(self):
+        metrics = super().generate_metrics()
+        for metric in metrics:
+            metric.labels.update(allocation_type=AllocationType.SHARES)
+            metric.name = 'allocation_%s' % AllocationType.SHARES.value
```
pskrzyns: could you use new style string formatting (str.format) ?

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
             cpu_usage = int(cpu_usage_file.read())
 
         return {MetricName.CPU_USAGE_PER_TASK: cpu_usage}
+
+    def _read(self, cgroup_control_file: str) -> int:
+        """Read helper to store any and convert value from cgroup control file."""
+        with open(os.path.join(self.cgroup_fullpath, cgroup_control_file)) as f:
+            v = int(f.read())
```
pskrzyns: could you use meaningful names compatible with pep8 ?
how will be exception handled i.e. not existing file ?

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+
+        self._write(CPU_SHARES, shares)
+
+    def get_normalized_quota(self) -> float:
+        """Read normlalized quota against configured period and number of available cpus."""
+        assert self.allocation_configuration is not None, \
+            'normalization configuration cannot be used without configuration!'
+        current_quota = self._read(CPU_QUOTA)
+        current_period = self._read(CPU_PERIOD)
+        if current_quota == -1:
```
pskrzyns: Could you use constants instead of values ?

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+        current_period = self._read(CPU_PERIOD)
+
+        if current_period != self.allocation_configuration.cpu_quota_period:
+            self._write(CPU_PERIOD, self.allocation_configuration.cpu_quota_period)
+
+        if quota_normalized >= 1.0:
+            quota = -1
+        else:
+            # synchornize period if nessesary
+            quota = int(quota_normalized * self.allocation_configuration.cpu_quota_period *
```
pskrzyns: maybe 
```quota = min([1000, int(quota_normalized * self.allocation_configuration.cpu_quota_period * self.platform_cpus)])```

Comment (ppalucki) [reply owca/cgroups.py]: first its a bug, you should to use max() to guarantee that at least "1000" is set, 
second (because even you proposed it wrong) it is too magical - `if x < MIMNIUM: x= MINIMUME` 
(explicit is better than implict) 
and you didn't use constant for "1000" :)

File: owca/cgroups.py (hunk 3 above)
Comment (ppalucki) [reply owca/cgroups.py]: is should fail, the preconditions of exiting cgroups and resctrl filesystems are checked at the startup 

general rule of thumb: do not handle exception unless you don't know what to do with them :) but let it fire

Comment (ppalucki) [reply owca/cgroups.py]: your're right about naming - will fix that

File: owca/cgroup_allocations.py (hunk 2 above)
Comment (ppalucki) [reply owca/cgroup_allocations.py]: most of our code uses old_style formating for simple reason: because it is compatible with 
how loggers formating works (also uses old style formating)

IMO it's better to keep the formating method consisten across the project

File: owca/cgroups.py (hunk 1 above)
Comment (ppalucki) [reply owca/cgroups.py]: you're right about naming - I will fix it 
check may response below, regarding quota - I believe that 'if' condition with comment is more explicit and better

Comment (ppalucki) [reply owca/cgroups.py]: @pskrzyns Thank for your feedback !

File: owca/cgroups.py (hunk 4 above)
Comment (ppalucki) [reply owca/cgroups.py]: of course - will do !

File: owca/cgroups.py [hunk 6]
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+    def _read(self, cgroup_control_file: str) -> int:
+        """Read helper to store any and convert value from cgroup control file."""
+        with open(os.path.join(self.cgroup_fullpath, cgroup_control_file)) as f:
+            v = int(f.read())
+            log.log(logger.TRACE, 'cgroup: read %s=%r', f.name, v)
+            return v
+
+    def _write(self, cgroup_control_file: str, value: int):
+        """Write helper to store any int value in cgroup control file."""
+        with open(os.path.join(self.cgroup_fullpath, cgroup_control_file), 'wb') as f:
```
iwankgb: s/f/cgroup/

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+        self._write(CPU_SHARES, shares)
+
+    def get_normalized_quota(self) -> float:
+        """Read normlalized quota against configured period and number of available cpus."""
+        assert self.allocation_configuration is not None, \
+            'normalization configuration cannot be used without configuration!'
+        current_quota = self._read(CPU_QUOTA)
+        current_period = self._read(CPU_PERIOD)
+        if current_quota == -1:
+            return 1.0
```
iwankgb: Another magic number. Use a constant, please.

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+    def set_normalized_quota(self, quota_normalized: float):
+        """Unconditionally sets quota and period if nessesary."""
+        assert self.allocation_configuration is not None, \
+            'setting quota cannot be used without configuration!'
+        current_period = self._read(CPU_PERIOD)
+
+        if current_period != self.allocation_configuration.cpu_quota_period:
+            self._write(CPU_PERIOD, self.allocation_configuration.cpu_quota_period)
+
+        if quota_normalized >= 1.0:
```
iwankgb: Replace magin number with constant, please.

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+            quota = -1
```
iwankgb: Ditto.

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+    def get_allocations(self) -> TaskAllocations:
+        assert self.allocation_configuration is not None, \
+            'reading normalized allocations is not possible without configuration!'
+        return {
+            AllocationType.QUOTA: self.get_normalized_quota(),
+            AllocationType.SHARES: self._get_normalized_shares(),
+        }
+
+    def get_pids(self) -> List[str]:
+        with open(os.path.join(self.cgroup_fullpath, TASKS)) as f:
```
iwankgb: s/f/cgroup/

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+        assert self.allocation_configuration is not None, \
+            'normalization configuration cannot be used without configuration!'
+        current_quota = self._read(CPU_QUOTA)
+        current_period = self._read(CPU_PERIOD)
+        if current_quota == -1:
+            return 1.0
+        # Period 0 is invalid arugment for cgroup cpu subsystem. so division is safe.
+        return current_quota / current_period / self.platform_cpus
+
+    def set_normalized_quota(self, quota_normalized: float):
```
iwankgb: Written value is not normalized - you may set `-1` to disable quota. What do you think about following function signature:
```python
def set_quota(normalized_quota: float)
```
It makes perfectly clear that normalized value is accepted on input but does not suggest that value written to cgroup fs is normalized.

owca/cgroups.py
```diff
@@ -40,3 +52,80 @@ def get_measurements(self) -> Measurements:
...
+            f.write(raw_value)
+
+    def _get_normalized_shares(self) -> float:
+        """Return normalized using cpu_shreas_min and cpu_shares_unit for normalization."""
+        assert self.allocation_configuration is not None, \
+            'normalization configuration cannot be used without configuration!'
+        shares = self._read(CPU_SHARES)
+        return (shares / self.allocation_configuration.cpu_shares_unit)
+
+    def set_normalized_shares(self, shares_normalized):
```
iwankgb: Change function signature to:
```python
def set_shares(self, normalized_shares: float):
```
(See my comment to quota sharing for explanation)

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+    ctrl_file_mock = MagicMock()
+    full_path = '/sys/fs/cgroup/cpu/some/foo1/some_ctrl_file'
+    open_mock = create_open_mock({full_path: ctrl_file_mock})
+    with patch('builtins.open', open_mock):
+        cgroup._write('some_ctrl_file', 5)
+    ctrl_file_mock.assert_called_once_with(full_path, 'wb')
+    ctrl_file_mock.assert_has_calls([call().__enter__().write(b'5')])
+
+
+@patch('owca.containers.Cgroup._read', return_value=1000)
```
iwankgb: I have to admin that I'm not particularly happy when tested class is being mocked. It looks like violation of SRP to me: `Cgroup` responsibilities are: normalization, denormalization, writing to fs and reading from fs. (De)normalization has nothing to do with writing and reading.
We should have a quick chat about it today or tomorrow, I think.

iwankgb: What will happen if you delete `_read()` method from `Cgroup` class?

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+                                cpu_quota_period=cpu_quota_period))
+            cgroup.set_normalized_quota(normalized_quota)
+            write_mock.assert_has_calls([call('cpu.cfs_quota_us', expected_quota_write)])
+            if expected_period_write:
+                write_mock.assert_has_calls([call('cpu.cfs_period_us', expected_period_write)])
+
+
+@patch('owca.containers.PerfCounters')
+@patch('owca.containers.Cgroup')
+def test_cgroup_allocations(Cgroup_mock, PerfCounters_mock):
```
iwankgb: This function should be moved to `test_cgroup_allocations.py`

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+             None, -1),
+            (1., 1000, 8, 1000,
+             None, -1),
+            (.5, 1000, 8, 1000,
+             None, 4000),
+            (.25, 10000, 8, 1000,
+             None, 20000),
+    ]
+)
+def test_set_normalized_quota(normalized_quota, cpu_quota_period, platforms_cpu,
```
iwankgb:
```suggestion
def test_setting_quota(normalized_quota, cpu_quota_period, platforms_cpu,
```

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+
+@pytest.mark.parametrize(
+    'normalized_shares, allocation_configuration, expected_shares_write', [
+        (0., AllocationConfiguration(), 2),
+        (1., AllocationConfiguration(), 1000),  # based on cpu_shares_unit (default 1000)
+        (1., AllocationConfiguration(cpu_shares_unit=10000), 10000),
+        (2., AllocationConfiguration(cpu_shares_unit=10000), 20000),
+    ]
+)
+def test_set_normalized_shares(normalized_shares, allocation_configuration, expected_shares_write):
```
iwankgb:
```suggestion
def test_setting_shares(normalized_shares, allocation_configuration, expected_shares_write):
```

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+    cgroup = Cgroup('/some/foo1', platform_cpus=1,
+                    allocation_configuration=AllocationConfiguration())
+    assert cgroup.get_normalized_quota() == 1.0
+
+
+@patch('builtins.open', create_open_mock({
+    '/sys/fs/cgroup/cpu/some/foo1/tasks': '101\n102',
+    '/sys/fs/cgroup/cpu/foo2/tasks': '',
+}))
+def test_cgroup_get_pids():
```
iwankgb:
```suggestion
def test_getting_cgroup_pids():
```

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+    cgroup = Cgroup('/some/foo1', platform_cpus=1,
+                    allocation_configuration=AllocationConfiguration())
+    assert cgroup._get_normalized_shares() == pytest.approx(1, 0.01)
+
+
+@patch('builtins.open', create_open_mock({
+    '/sys/fs/cgroup/cpu/some/foo1/cpu.cfs_period_us': '100000',
+    '/sys/fs/cgroup/cpu/some/foo1/cpu.cfs_quota_us': '-1',
+}))
+def test_get_normalized_quota():
```
iwankgb:
```suggestion
def test_getting_normalized_quota():
```

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+    full_path = '/sys/fs/cgroup/cpu/some/foo1/some_ctrl_file'
+    open_mock = create_open_mock({full_path: ctrl_file_mock})
+    with patch('builtins.open', open_mock):
+        cgroup._write('some_ctrl_file', 5)
+    ctrl_file_mock.assert_called_once_with(full_path, 'wb')
+    ctrl_file_mock.assert_has_calls([call().__enter__().write(b'5')])
+
+
+@patch('owca.containers.Cgroup._read', return_value=1000)
+def test_get_normalized_shares(_read_mock):
```
iwankgb:
```suggestion
def test_getting_normalized_shares(_read_mock):
```

tests/test_cgroups.py
```diff
@@ -12,16 +12,135 @@
...
+
+
+@patch('builtins.open', mock_open(read_data='100'))
+def test_cgroup_read():
+    cgroup = Cgroup('/some/foo1', platform_cpus=1)
+    value = cgroup._read('some_ctr_file')
+    assert value == 100
+
+
+def test_cgroup_write():
```
iwankgb: This function and `test_cgroup_read()` look like test of another set of functionality to me. To be discussed along with my another comment.

owca/cgroups.py
ppalucki: fixed - renamed to file (it is not a cgroup) it its control_file like cpu.shares

owca/cgroups.py
ppalucki: fixed

owca/cgroups.py
ppalucki: fixed

tests/test_cgroups.py
ppalucki: agreed offline to follow convention of naming test_ functions with name of tested functions

e.g.
```python
def do_something:
  ... 
class A:
  def ble_method():
    ...
```
test will be
```python
def test_do_something:
   ...
def test_A_ble_method:
   ...
```

ppalucki: fixed

tests/test_cgroups.py
ppalucki: ditto - previous answers

tests/test_cgroups.py
ppalucki: ditto - previous answers

PR #49, opened 2019-02-14
Replace improper Task abstract class with dataclass

ppalucki:
Rationale:
- **abstractproperty** were used improperly 
- **abstractproperty** can only enforce method based "property" not the normal one, that is way **MesosTask** couldn't properly iherit from it

Now Task dataclass is enough and Node implementation have no create owns Task implementation and can just reuse existing dataclass.

Quirks:
- in MesosTask we still have to create __hash__ classes because how dataclass treat __hash__ method ([here is explanation](https://github.com/python/cpython/blob/3.7/Lib/dataclasses.py#L118))
- base Task class cannot have default properties (e.g. labels or resources) - because then all inherited class need to have all fields optional (check [this ](https://stackoverflow.com/questions/51575931/class-inheritance-in-python-3-7-dataclasses)fur further discussion)

### DIFF
--- a/owca/mesos.py
+++ b/owca/mesos.py
@@ -13,18 +13,16 @@
 # limitations under the License.
 
 
-from dataclasses import dataclass, field
-from typing import Dict, List, Union
-import urllib.parse
 import logging
 import os
+import urllib.parse
+from typing import List, Union
 
 import requests
+from dataclasses import dataclass
 
-# Mesos tasks id
-from owca.nodes import Node, TaskId
 from owca.metrics import Measurements, Metric
-
+from owca.nodes import Node, Task
 
 MESOS_TASK_STATE_RUNNING = 'TASK_RUNNING'
 CGROUP_DEFAULT_SUBSYSTEM = 'cpu'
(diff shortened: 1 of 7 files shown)


owca/mesos.py
```diff
@@ -33,30 +31,18 @@
...
-    cgroup_path: str  # Starts with leading "/"
-    labels: Dict[str, str] = field(default_factory=dict)
-
-    # Resources assigned accorind Mesos/Aurora task definition.
-    resources: Dict[str, float] = field(default_factory=dict)
-
     def __hash__(self):
-        """Every instance of mesos task is uniqully identified by cgroup_path.
-        Assumption here is that every mesos task is represented by one main cgroup.
+        """Reuse __hash__ method from base class, because dataclasses by default
```
iwankgb: I think that your dosctring is incomplete.

owca/mesos.py
```diff
@@ -33,30 +31,18 @@
...
     agent_id: str
 
-    # inferred
-    cgroup_path: str  # Starts with leading "/"
-    labels: Dict[str, str] = field(default_factory=dict)
-
-    # Resources assigned accorind Mesos/Aurora task definition.
-    resources: Dict[str, float] = field(default_factory=dict)
-
     def __hash__(self):
```
iwankgb: Can we use `__eq__` instead?

ppalucki: no we cannot: task is used as key in container_manager and __eq__ is not enough

owca/testing.py
```diff
@@ -98,14 +98,10 @@ def anomaly(contended_task_id: TaskId, contending_task_ids: List[TaskId],
 
 def task(cgroup_path, labels=None, resources=None):
     """Helper method to create task with default values."""
-    return MesosTask(
+    return Task(
         cgroup_path=cgroup_path,
         name='name-' + cgroup_path,
-        executor_pid=1,
-        container_id='container_id-' + cgroup_path,
         task_id='task-id-' + cgroup_path,
```
felidadae: Not to break tests
```suggestion
        task_id= cgroup_path[1:] + 'task-id',
```

PR #50, opened 2019-02-15
Validation: Config files containers kinds (list/dict) type fields validation

ppalucki:
Validate recursively proper types of containers types like dict, list and basic types.


owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+
+class WeekValidationError(Exception):
+    """Used for weak validation errors, that handling depends on strict_mode setting.
+    Such errors can be handled in following ways:
+    - re-raise ValidationError exception
+    - just print log.warning.
+    """
+
+
+def _validate_value(value, expected_type):
```
iwankgb: It seems to be the most complex function in the whole project:
```
(owca) ➜  owca git:(fields-validation) ✗ flake8 --max-complexity=15 owca
owca/config.py:65:1: C901 '_validate_value' is too complex (19)
```
I believe that unit tests must be provided to assure it works as expected.

ppalucki: surprising it is less complex that previous solution - but you're right, it requires own tests

ppalucki: will do

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
     pass
 
 
-def _constructor(loader: yaml.loader.Loader, node: yaml.nodes.Node, cls: type):
+class ValidationError(Exception):
+    """Used for serious validation errors, that force program to stop. """
+
+
+class WeekValidationError(Exception):
```
iwankgb: s/Week/Weak/

owca/config.py
ppalucki: @iwankgb done, please re review

ppalucki: tests added here: https://github.com/intel/owca/pull/50/files#diff-e5c1a6b5f47b6a145fd20e8cb7aef9f7

owca/config.py
ppalucki: fixed

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+
+
+def _validate_value(value, expected_type):
+    """Should raise if value is not an instance of expected_type.
+
+    """
+    # Limit checking to only first level of generic parametrized types.
+    # so we can check List[int] and List[List] but no List[int]
+
+    if expected_type == inspect._empty:
```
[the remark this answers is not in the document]
iwankgb: What about:
```python
from inspect import Parameter
if expected_type == Parameter.empty
```
As far as I understand `Parameter` class represents values passed as function arguments.

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+                            from e
+
+                return
+
+        elif issubclass(expected_type, typing.Dict):
+            if not isinstance(value, dict):
+                raise ValidationError('expected dict-like type %r, but got %r', expected_type,
+                                      type(value))
+            else:
+                expected_key_type, expected_sub_value_type = expected_type.__args__
```
iwankgb: s/expected_sub_value_type/expected_value_type/
It would be more consistent when we talk about dicts.

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+                for key, sub_value in value.items():
```
iwankgb: s/key/item_key/
s/sub_value/item_value/

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+                    try:
+                        _validate_value(key, expected_key_type)
```
iwankgb: I think that `_validate_value` needs to be renamed. What about `_assure_type()`? It is weird when you see a call to `_validate_value()` function and first variable pass is called `key`.

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+            if isinstance(value, union_expected_type):
+                break
+        else:
+            # Value doesn't match any type from union.
+            raise ValidationError(
+                'improper type from union (got=%r expected=%r)!' % (type(value), expected_type))
+
+    # Handle simple types.
+    else:
+        assert not isinstance(value, typing.GenericMeta), 'should be handled by cases above !!!'
```
iwankgb: What is this assertion for?

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+    """
+    # Limit checking to only first level of generic parametrized types.
+    # so we can check List[int] and List[List] but no List[int]
+
+    if expected_type == inspect._empty:
+        raise WeakValidationError('missing type declaration!')
+
+    if isinstance(expected_type, typing.GenericMeta):
+        if issubclass(expected_type, typing.List):
+            if not isinstance(value, list):
```
iwankgb: Consider moving logic responsible for handling each type to separate function, eg:
```python
def _assert_is_list(value):
    (...)

def _assert_is_dict(value):
    (...)

def _assert_is_union(value):
    (...)

# etc, etc
```
It will increase readability and make some comments obsolete.

owca/config.py
```diff
@@ -165,14 +240,33 @@ def register(cls):
         foo = Foo.__new__(Foo)  # construct uninitialized (blank) instance of given class
         foo.__init__(x=1) # initialize instance
 
+
+    Can be used both as function, decorator or paremetrized decorator.
```
iwankgb: How is "decorator" different from "parametrized decorator"? I think we can drop last alternative from this line but leave it as an example below.

owca/config.py
```diff
@@ -165,14 +240,33 @@ def register(cls):
...
+
+    @register  # with default strict_mode=True)
```
iwankgb:
```suggestion
    @register  # with default strict_mode=True
```

owca/config.py
```diff
@@ -165,14 +240,33 @@ def register(cls):
...
+        # Just simply register new constructor for given cls.
+        log.log(logger.TRACE, 'registered class %r' % (cls.__name__))
+        _yaml.constructor.add_constructor(
+            '!%s' % cls.__name__, functools.partial(_constructor, cls=cls, strict_mode=strict_mode))
+        _registered_tags.add(cls.__name__)
+        return cls
+
+    # See if we're being called as @dataclass or @dataclass().
+    if cls is None:
+        # We're called with parens with decorator parameters e.g. strict_mode
```
iwankgb: I can't understand what you mean. Could you rephrase it, please? Why `@dataclass` above?

tests/config/test_validate_values.py
```diff
@@ -0,0 +1,71 @@
...
+])
+def test_validate_value_good(value, expected_type):
+    _validate_value(value, expected_type)
+
+
+@pytest.mark.parametrize('value, expected_type, expected_exception_msg', [
+    (1, float, 'int'),
+    (1.2, int, 'float'),
+    (True, float, 'bool'),
+    (2.5, Optional[bool], 'Union'),
```
iwankgb: Why a `Union`? Could you add a comment explaining this case, please?

iwankgb:
In #57 there is an `Enum` used in configuration. Will your validation handle such case?

tests/config/test_validate_values.py
ppalucki: `Opitional[x]` is just `Union[x, NoneType]`

added a comment

owca/config.py
```diff
@@ -165,14 +250,33 @@ def register(cls):
         foo = Foo.__new__(Foo)  # construct uninitialized (blank) instance of given class
         foo.__init__(x=1) # initialize instance
 
+
+    Can be used both as function, decorator or parametrized decorator:
```
iwankgb:
```suggestion
    Can be used as function or decorator.
```
As far as I understand you can pass arguments to any decorator that accepts arguments.

owca/config.py
```diff
@@ -51,7 +50,88 @@ class ConfigLoadError(Exception):
...
+
+def _validate_value(value, expected_type):
+    """Should raise if value is not an instance of expected_type.
+
+    """
+    # Limit checking to only first level of generic parametrized types.
+    # so we can check List[int] and List[List] but no List[int]
+
+    if expected_type == inspect._empty:
+        raise WeakValidationError('missing type declaration!')
```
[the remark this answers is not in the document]
ppalucki: agreed on stand up meeting that we want this functionality

owca/config.py
ppalucki: we don't throw away strict_mode and this was explained offline

ppalucki:
@iwankgb added a tests that Enum is handled properly

owca/config.py
ppalucki: fixed - it is now `_assure_type` or `_assure_SOME_type`

tests/config/test_dataclasses_validation_invalid_list.yaml
```diff
@@ -0,0 +1,20 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+
+dc1: !DCExampleValidation
+  i: [5]
+  rr:
+    - [2, 3]
+    - ['asdf']
```
felidadae: for me it looks weird `asdf`, could you replace that random string with `random_string` or `a`

tests/config/test_dataclasses.py
```diff
@@ -11,24 +11,29 @@
...
+    BAR = 'barek'
+    BAZ = 'bazek'
+
+
 @config.register
 @dataclass
 class DCExample:
     x: int
-    y: str = 'bleblo'
+    y: Optional[str] = 'bleblo'
```
felidadae: please no `bleblo`, i could maybe accept `bleblorian`, but prefer much more than anything like that just `a` or `abc` or something more predictable

tests/config/test_dataclasses_validation_invalid_list.yaml
ppalucki: fixed

PR #52, opened 2019-02-20
Handle task's pid races, during cgroups.tasks and resctrl sync

ppalucki:
In specific environments when process pids changing rapidly, owca encounrs race conditions because pids is not valid anymore 

if was handled improperly this PR fixes that


owca/resctrl.py
```diff
@@ -130,31 +129,44 @@ def sync(self, max_attempts=3):
...
-                    'sync: Succesful synchronization for %r - braking' % self.resgroup_dir)
-            break
-
-        else:
+            log.log(logger.TRACE, 'sync: Writings tasks for %r' % (self.resgroup_dir))
+            with SetEffectiveRootUid():
+                try:
+                    f = open(self.resgroup_tasks, 'w')
+                    for task in tasks.split():
+                        # is that my code!!!
```
iwankgb: Do you think something is wrong with the code below? ;)

owca/resctrl.py
```diff
@@ -130,31 +129,44 @@ def sync(self, max_attempts=3):
...
+                    try:
+                        f.close()
+                    except OSError:
+                        # Note: handle/fd will be released automatically after returning from
+                        # function!
+                        log.warning('Could not close file - ignoring!')
+
+            # In any case other than successful flush, retry ...
+            attempt += 1
+            continue
```
iwankgb: Is this statement necessary? This is last line of the loop so it is redundant in my opinion.

owca/resctrl.py
ppalucki: fixed - good catch !

owca/resctrl.py
ppalucki: again your're completly right - my mistake - removed

PR #54, opened 2019-02-26
Allocation[5]: containers & platforms

ppalucki:


owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+    def cleanup(self):
+        # _cleanup
+        for container in self.containers.values():
+            container.cleanup()
+
+
+def _calculate_desired_state(
+        discovered_tasks: List[Task], known_containers: List[Container]
+) -> (List[Task], List[Container]):
+    """Prepare desired state of system by comparing actual running Mesos tasks and already
```
iwankgb: This is not true.

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+    watched containers.
+
+    Assumptions:
+    * One-to-one relationship between task and container
+    * cgroup_path for task and container need to be identical to establish the relationship
+    * cgroup_path is unique for each task
+
+    :returns "list of Mesos tasks to start watching"
+    and "orphaned containers to _cleanup" (there are no more Mesos tasks matching those containers)
```
iwankgb:
```suggestion
    and "orphaned containers to clean up" (there are no more Mesos tasks matching those containers)
```

owca/platforms.py
```diff
@@ -222,10 +230,46 @@ def read_cpu_info_file(cpu, rel_file_path) -> int:
     return nr_of_online_cpus, nr_of_cores, nr_of_sockets
 
 
-def collect_platform_information() -> (Platform, List[Metric], Dict[str, str]):
-    """Returns Platform information, metrics and common labels.
+def collect_rdt_information(rdt_enabled: bool) -> (
+        Optional[str],  # cbm_mask
```
iwankgb: No one is going to understand *6* returned values. Use a class instead.

owca/platforms.py
```diff
@@ -222,10 +230,46 @@ def read_cpu_info_file(cpu, rel_file_path) -> int:
...
+
+        return (cbm_mask, min_cbm_bits, rdt_mb_control_enabled, num_closids,
+                mb_bandwidth_gran, mb_min_bandwidth)
+    else:
+        return None, None, False, None, None, None
 
 
+def collect_platform_information(rdt_enabled: bool = True) -> (
+        Platform, List[Metric], Dict[str, str]):
+    """Returns Platform information, metrics and common labels.
```
iwankgb: `Dict[str, str]` should be explained below as `Platform` and `List[Metric]` are.

owca/platforms.py
```diff
@@ -235,10 +279,27 @@ def collect_platform_information() -> (Platform, List[Metric], Dict[str, str]):
...
-    nr_of_cpus, nr_of_cores, nr_of_sockets = collect_topology_information()
-    platform = Platform(sockets=nr_of_sockets, cores=nr_of_cores, cpus=nr_of_cpus,
-                        cpus_usage=parse_proc_stat(read_proc_stat()),
-                        total_memory_used=parse_proc_meminfo(read_proc_meminfo()),
-                        timestamp=time.time())
-    assert len(platform.cpus_usage) == platform.cpus, "Inconsistency in cpu data returned by kernel"
+    # Static information
+    nr_of_cpus, nr_of_cores, no_of_sockets = collect_topology_information()
+    (rdt_cbm_mask, rdt_min_cbm_bits, rdt_mb_control_enabled, num_closids,
+     rdt_mb_bandwidth_gran, rdt_mb_min_bandwidth) = collect_rdt_information(rdt_enabled)
```
iwankgb: This is unreadable.

owca/platforms.py
```diff
@@ -235,10 +279,27 @@ def collect_platform_information() -> (Platform, List[Metric], Dict[str, str]):
...
+    # Dynamic information
+    cpus_usage = parse_proc_stat(read_proc_stat())
+    total_memory_used = parse_proc_meminfo(read_proc_meminfo())
+    platform = Platform(sockets=no_of_sockets,
+                        cores=nr_of_cores,
+                        cpus=nr_of_cpus,
+                        cpus_usage=cpus_usage,
+                        total_memory_used=total_memory_used,
+                        timestamp=time.time(),
+                        rdt_mb_control_enabled=rdt_mb_control_enabled,
```
iwankgb: What about single `Platform.rdt` property storing all RDT-related information?

tests/test_containers.py
```diff
@@ -0,0 +1,110 @@
...
+        ([task('/t2')], [container('/t1'), container('/t2')],  # first task just disappeared
+         [], [container('/t1')]),  # remove the first container
+        # some other cases
+        ([task('/t1'), task('/t2')], [],  # the new task, just appeared
+         [task('/t1'), task('/t2')], []),
+        ([task('/t1'), task('/t3')], [container('/t1'),
+                                      container('/t2')],  # t2 replaced with t3
+         [task('/t3')], [container('/t2')]),  # nothing to do,
+    ))
+def test_calculate_desired_state(
```
iwankgb: Desired state it is not!

owca/platforms.py
jellonek: ... or https://docs.python.org/3/library/collections.html#collections.namedtuple

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
     return all_measurements_flat
 
 
+def _convert_cgroup_path_to_resgroup_name(cgroup_path):
+    """Return resgroup compatbile name for cgroup path (remove special characters like /)."""
```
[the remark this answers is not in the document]
ppalucki: What is wrong with function name ?
Function name says what is a purpose (cgroup_path - converted to somthing acceptable as resgroup name)
What is your suggestion ? 

I can drop docstring. of course.

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
 
     def __post_init__(self):
-        self.cgroup = Cgroup(self.cgroup_path)
+        self.cgroup = Cgroup(
+            self.cgroup_path,
+            platform_cpus=self.platform_cpus,
+            allocation_configuration=self.allocation_configuration,
+        )
+        self.container_name = (self.container_name or
+                               _convert_cgroup_path_to_resgroup_name(self.cgroup_path))
```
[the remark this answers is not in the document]
ppalucki: Ok, but to what, what is wrong with the name of function?

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+                  self.container_name, self.resgroup, pprint.pformat(allocations))
+
+        return allocations
+
+
+class ContainerManager:
+    """Main engine of synchornizing state between found orechestratios software tasks,
+    its containers and resctrl system.
+
+    - sync_container_state - is responsible for mapping Tasks to Container objects
```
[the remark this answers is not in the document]
ppalucki: will do - it was already outdated ! :)

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+        and their external resources, e.g.:
+        - perf counters opens file descriptors for counters,
+        - resctrl (ResGroups) creates and manages directories under resctrl fs and scarce "clsid"
+            hardware identifiers
+
+        Can throw OutOfClosidsException.
+        """
+
+        # Find difference between discovered Mesos tasks and already watched containers.
+        new_tasks, containers_to_cleanup = _calculate_desired_state(
```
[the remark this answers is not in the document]
ppalucki: Sounds better, done.

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+            tasks, list(self.containers.values()))
+
+        if containers_to_cleanup:
+            log.debug('sync_containers_state: cleaning up %d containers',
+                      len(containers_to_cleanup))
+            log.log(logger.TRACE, 'sync_containers_state: containers_to_cleanup=%r',
+                    containers_to_cleanup)
+
+        # Cleanup and remove orphaned containers (_cleanup).
```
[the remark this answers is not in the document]
ppalucki: I do not understand - "calculcate_desired_state" (now `find_new_and_dead_tasks`) was about what we want accomplish. and here we are taking necessary actions  to change the current state closer to desired one. 
So what is wrong with ?
- calculate_desired_state: "return orphaned containers to deleted"
- here: "cleanup and remove orphaned contianers"
?

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+        for container_to_cleanup in containers_to_cleanup:
```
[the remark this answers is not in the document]
ppalucki: it doesn't matter because looping over empty list do nothing, 
condition was added only to disable logs with empty list

moved for clarity

it is un-pythonic to:
```python
if list:
  for i in list:
    do(i)
```
but with comments is acceptable

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+                           for task, container in self.containers.items()
+                           if task in tasks}
+
+        if new_tasks:
+            log.debug('sync_containers_state: found %d new tasks', len(new_tasks))
+            log.log(logger.TRACE, 'sync_containers_state: new_tasks=%r', new_tasks)
+
+        # Prepare state of currently assigned resgroups
+        # and remove some orphaned resgroups
+        container_name_to_ctrl_group = {}
```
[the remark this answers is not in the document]
ppalucki: done - wasn't sure about scope of local block variable and for clarity, wanted to have to out of block to be accessible from anothre block

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+        if self.rdt_enabled:
+            mon_groups_relation = resctrl.read_mon_groups_relation()
+            log.debug('mon_groups_relation: %s', pprint.pformat(mon_groups_relation))
+            resctrl.clean_taskless_groups(mon_groups_relation)
+
+            mon_groups_relation = resctrl.read_mon_groups_relation()
```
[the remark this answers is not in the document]
ppalucki: it is not only for logging, we need that information to reestablish container and resgroup relation - is is used to create inverse relation `container_name_to_ctrl_group` which represents what relation is actually used in resctrl.

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+            )
+            self.containers[new_task] = container
+
+        # Sync "state" of individual containers.
+        # Note: only the pids are synchronized, not the allocations.
+        for container in self.containers.values():
+            if self.rdt_enabled:
+                if container.container_name in container_name_to_ctrl_group:
+                    resgroup_name = container_name_to_ctrl_group[container.container_name]
+                    container.resgroup = ResGroup(name=resgroup_name)
```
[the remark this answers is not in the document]
ppalucki: `ResGroup` has no state - it is only the "pointer" to `/sys/fs/resctrl/GROUP' - so I want to make sure that every iteration we operating on fresh data

it is required for restarting owca that state is only in filesystem

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+                else:
+                    # Every newly detected containers is first assigne to root group.
+                    container.resgroup = ResGroup(name='')
+            container.sync()
+
+        return self.containers
+
+    def cleanup(self):
+        # _cleanup
```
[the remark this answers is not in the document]
ppalucki: done

owca/containers.py
ppalucki: I do not understand ? Could you elaborate - docstring was changed and now 
it is:
```python
def _find_new_and_dead_tasks(
        discovered_tasks: List[Task], known_containers: List[Container]
) -> (List[Task], List[Container]):
    """Return the list of newly discovered and containers without tasks,
    by comparing actually running tasks and already watched containers.

```

owca/platforms.py
ppalucki: @jellonek are you still programming python ? :) now we have "nametuples on steroids" called "[dataclasses](https://docs.python.org/3/library/dataclasses.html)" - like a simple structs in go :)

will do

owca/platforms.py
ppalucki: done

owca/platforms.py
ppalucki: fixed - we have now `RDTInformation` dataclass

owca/platforms.py
ppalucki: I don't want to change API again for SSG PRC ? Could you ask them what they think tomorrow? I they don't object I can change that (is was find for one propery and now it is horrible for 5)

tests/test_containers.py
ppalucki: change the name of the funciton - now it is test_find_new_and_dead_tasks

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
+            # Calculate inverse relastion of container_name
+            # to res_group name based on mon_groups_relations
+            for ctrl_group, container_names in mon_groups_relation.items():
+                for container_name in container_names:
+                    container_name_to_ctrl_group[container_name] = ctrl_group
+            log.debug('container_name_to_ctrl_group: %s',
+                      pprint.pformat(container_name_to_ctrl_group))
+
+        # Create new containers and store them.
+        for new_task in new_tasks:
```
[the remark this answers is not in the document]
ppalucki: it is the same situtation as before

for on empty list is equivalent to if "empty list" - nobody guards looping with if it is no required

owca/containers.py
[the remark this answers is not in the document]
iwankgb: In this very case it is used to generate `container_name` attribute. How is container name related to resctrl group name? The function responsibility is to derive unique identifier from cgroup path and the identifier is used as:
* container name
* name of directory in resctrl filesystem.

owca/containers.py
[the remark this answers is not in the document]
iwankgb: See [another thread](https://github.com/intel/owca/pull/54/files#r260233598), please. Function is not public and I don't think it needs a docstring if name reflects what we use it for.

owca/containers.py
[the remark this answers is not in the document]
iwankgb: Ignore, please. It looks like myself misunderstanding something ;)

owca/containers.py
```diff
@@ -39,36 +44,200 @@ def flatten_measurements(measurements: List[Measurements]):
...
+
+    def cleanup(self):
+        for container in self.containers.values():
+            container.cleanup()
+
+
+def _find_new_and_dead_tasks(
+        discovered_tasks: List[Task], known_containers: List[Container]
+) -> (List[Task], List[Container]):
+    """Return the list of newly discovered and containers without tasks,
```
iwankgb:
```suggestion
    """Returns the of newly discovered tasks and containers without tasks,
```

owca/containers.py
```diff
@@ -39,36 +44,200 @@ def flatten_measurements(measurements: List[Measurements]):
...
+    by comparing actually running tasks and already watched containers.
```
iwankgb:
```suggestion
    by comparing running tasks against list of known containers.
```

owca/platforms.py
```diff
@@ -230,40 +230,40 @@ def read_cpu_info_file(cpu, rel_file_path) -> int:
...
+    num_closids: Optional[int]
+    mb_bandwidth_gran: Optional[int]
+    mb_min_bandwidth: Optional[int]
+
 
-        return (cbm_mask, min_cbm_bits, rdt_mb_control_enabled, num_closids,
-                mb_bandwidth_gran, mb_min_bandwidth)
+def collect_rdt_information(rdt_enabled: bool) -> RDTInformation:
+    """Returns rdt information values."""
+    with open('/sys/fs/resctrl/info/L3/cbm_mask') as f:
```
iwankgb: What happens if resctrl fs is not mounted?

iwankgb: `rdt_enabled` seems not to be used; why is this passed to the function?

owca/platforms.py
```diff
@@ -274,15 +274,18 @@ def collect_platform_information(rdt_enabled: bool = True) -> (
...
+        (generated by other sources) e.g. host
 
     Note: returned metrics should be consistent with information covered by platform
 
     """
     # Static information
     nr_of_cpus, nr_of_cores, no_of_sockets = collect_topology_information()
-    (rdt_cbm_mask, rdt_min_cbm_bits, rdt_mb_control_enabled, num_closids,
-     rdt_mb_bandwidth_gran, rdt_mb_min_bandwidth) = collect_rdt_information(rdt_enabled)
+    if rdt_enabled:
```
iwankgb: Shouldn't this logic be moved into `collect_rdt_information()`? It would prevent user from calling public function and having exception raised (see my comment to the function itself).

owca/platforms.py
iwankgb: We will raise the issue with them and ask what they think.

tests/test_containers.py
```diff
@@ -0,0 +1,110 @@
...
+        got_containers = containers_manager.sync_containers_state(tasks)
+
+    # Check internal state ...
+    assert sorted(got_containers) == sorted(expected_running_containers)
+
+    # Check other side effects like calling sync() on external objects.
+    assert sync_mock.call_count == len(expected_running_containers)
+
+    # Check container objects has proper resgroup assigned.
+    for resgroup_name, container_names in mon_groups_relation.items():
```
iwankgb: @intel/owca-write, I would like to discuss this piece of code. In my opinion it smells bad because there is an assert emedded in four control structures; it makes me think about testing this code snippet. Do you have any idea how we could simplify it?

owca/platforms.py
ppalucki: it was there initially but after adding RDTInformation I moved that out - and rdt_enabled is directive from user - not read from system

tests/test_containers.py
jellonek:
```python
gc = {g.container_name: g.resgroup.name for g in got_containers.values()}
for resgroup_name, container_names in mon_group_relation.items():
    for name in container_names:
        assert gc.get(name) == resgroup_name
```
... or something like that...

owca/platforms.py
jellonek:
```python
def _read_value(path, type_=None):
    with open(path) as f:
        data_as_str = f.read().strip()
    return type_(data_as_str) if type_ else data_as_str

...
    cbm_mask = _read_value('/sys/fs/resctrl/info/L3/cbm_mask')
    min_cbm_bits = _read_value('/sys/fs/resctrl/info/L3/min_cbm_bits')
    num_closids = _read_value('/sys/fs/resctrl/info/L3/num_closids', int)
...
    rdt_mb_control_enabled = 'MB' in _read_value('/sys/fs/resctrl/schemata')
...
```

owca/platforms.py
```diff
@@ -222,23 +230,79 @@ def read_cpu_info_file(cpu, rel_file_path) -> int:
...
+    nr_of_cpus, nr_of_cores, no_of_sockets = collect_topology_information()
+    if rdt_enabled:
+        rdt_information = collect_rdt_information(rdt_enabled)
+    else:
+        rdt_information = RDTInformation(None, None, False, 0, 0, 0)
+
+    # Dynamic information
+    cpus_usage = parse_proc_stat(read_proc_stat())
+    total_memory_used = parse_proc_meminfo(read_proc_meminfo())
+    platform = Platform(sockets=no_of_sockets,
```
jellonek: Please keep the same way of formatting in multiple places, preferably keeping the syntax as in 187 line.
Do that by reformatting this instantiation moving argument to next line (that will change the indentation).

Please take a look on https://github.com/ambv/black - the standard tool used by most python core developers for files autoformatting.

tests/test_containers.py
iwankgb: It allows us to get rid of conditional statement and I believe that is's a really neat idea :)

owca/containers.py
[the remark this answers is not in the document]
ppalucki: Ok I got it, but still don't know what is your proposed name of function ?

Function name is used to convert "cgroup path" (possible deep) into flat simple "resgroup name" which is used as the container name. 

Do you want to express the intention using the name of the function?

owca/platforms.py
jellonek: btw. if resctrlfs is not mounted this code with raise an exception, but that should for sure that should be handled on other level (perhaps during some initialization procedure which is using this code).

owca/containers.py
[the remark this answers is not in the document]
ppalucki: I dropped the the docstring? What is your proposition for function name? IMO it is ok, now.
Do you want to call it `_flatten` ?

owca/platforms.py
ppalucki: @iwankgb you right, I forgot to remove rdt_enable from parameter lists - fixed
@jellonek rdt_enabled is checked already during initialization an returns respective error

owca/containers.py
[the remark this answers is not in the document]
felidadae: def _sanitize_cgroup_path(cgroup_path)

owca/platforms.py
ppalucki: interesting I did exactly the same some other place, but was lazy here and just copy&paste over and over again :) 

fixed [here](https://github.com/intel/owca/pull/54/commits/cd4d679d66f8812664cbebe0c891a264c5b5e1be) but without _type

owca/containers.py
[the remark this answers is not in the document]
ppalucki: done

owca/containers.py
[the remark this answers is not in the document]
ppalucki: According @felidadae suggestion I changed name to `_sanitize_cgroup_path`

tests/test_containers.py
ppalucki: you have just refactored one of the loop in list comprehension :) it is just hack - there are still three loops there :) only the indentation has changed

If you think both it is more readable, lets have it like that.

But I still need if, to make check first if container with given name was even returned:
```python
    # Check container objects has proper resgroup assigned.
    got_container_resgroup_names = {c.container_name:
                                    c.resgroup.name for c in got_containers.values()}
    for expected_resgroup_name, container_names in mon_groups_relation.items():
        for container_name in container_names:
            if container_name in got_container_resgroup_names:
                got_resgroup_name = got_container_resgroup_names.get(container_name)
                assert got_resgroup_name == expected_resgroup_name
```

now looks like this :)

owca/platforms.py
ppalucki: I have tried black but i does too much magic - and sometimes the output looks much worse so you need tune it carefully. Too much trouble and easier to fix that once during review.

ppalucki: could you please, paste o code instead of refering to line, I found nothing interesting in line 187
```python
def collect_topology_information() -> (int, int, int):	
```
probably it has changed already

owca/platforms.py
```diff
@@ -61,6 +61,14 @@ class Platform:
     # [unix timestamp] Recorded timestamp of finishing data gathering (as returned from time.time)
     timestamp: float
 
+    # rdt information
```
felidadae: Why not keep just object of type RDTInformation instead of copying the fields?

ppalucki: because this is API that another group already uses - Iwan will confirm that they are ok, with change and we can refactor that

ppalucki: (the structure was added 2h to just not use long list of return values)

owca/containers.py
```diff
@@ -39,36 +44,199 @@ def flatten_measurements(measurements: List[Measurements]):
...
+            container.sync()
+
+        return self.containers
+
+    def cleanup(self):
+        for container in self.containers.values():
+            container.cleanup()
+
+
+def _find_new_and_dead_tasks(
```
felidadae: _find_new_and_dead_tasks ->_divide_into_new_and_dead_tasks

iwankgb: @felidadae some of them are not new and not dead.

tests/test_containers.py
iwankgb: > But I still need if, to make check first if container with given name was even returned
Why? I have to admit that I don't understand it: can container exists without resctrl group when RDT is enabled?

owca/platforms.py
iwankgb: As @jellonek stated: if resctrl filesystem is not mounted then `collect_rdt_information()` will raise an exception. When you pass `rdt_enabled_ as arcument to the function and move conditional statement too then you make it more robust in my opinion.

owca/containers.py
```diff
@@ -39,36 +44,205 @@ def flatten_measurements(measurements: List[Measurements]):
...
 @dataclass
 class Container:
-
     cgroup_path: str
+    platform_cpus: int
+    resgroup: ResGroup = None
+    allocation_configuration: Optional[AllocationConfiguration] = None
     rdt_enabled: bool = True
+    rdt_mb_control_enabled: bool = False
+    container_name: str = None  # defaults to flatten value of provided cgroup_path
```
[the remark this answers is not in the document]
ppalucki: is sanitanize_cgroup_path suggested by @felidadae, ok ?

owca/platforms.py
ppalucki: this function is just a helper not an platforms API (I should prefix that with _), the only API provided by platforms is `collect_platform_infromation` which requires rdt_enabled that is already checked regarding resctrl filesystem mounted - we can try make it robust by remember OWCA is not a general use wrapper over resctrl and platforms but agent with specified API for detect and allocate callback - many other functions and arch decisions are based on that.

tests/test_containers.py
ppalucki: sync_container_state tries to keep in sync information from three different sources:
- orchestrator - currently running Mesos tasks/Kuberenetes pods
- OS - resctrl assignment of containers and resgroups 
- internal - containers objects with underlying perf counters (fds)

The real problem is synchronizing the state is really susceptible to races.
We can see resctrl from previous run, trying to attach counters to tasks that already don't exists and naturally see new tasks that aren't yet under resgroups or perf countres.

Regarding the specific problem of resctrl. 
Every process in system belongs to some resgroup, usually it is root default group (name=''), so every newly detected tasks is assumed to belong to root default group unless the resctrl says otherwise. This allows restart agent without destroying all the resource configuration.

This exact assertion we're discussing, checks exactly that - if there is an existing mapping between container and resgroup read from resctrl filesystem, please take that into consideration during sync. The question is should we do that for every even known container or only for newly detected. I chose to sync every container every iteration. It costs nothing, but guarantees that we operation on current state of the system.

owca/containers.py
ppalucki: @felidadae and @iwankgb do we have an agreement here?

tests/test_containers.py
```diff
@@ -0,0 +1,112 @@
...
+
+    # Check other side effects like calling sync() on external objects.
+    assert sync_mock.call_count == len(expected_running_containers)
+
+    # Check container objects has proper resgroup assigned.
+    got_container_resgroup_names = {c.container_name:
+                                    c.resgroup.name for c in got_containers.values()}
+    for expected_resgroup_name, container_names in mon_groups_relation.items():
+        for container_name in container_names:
+            if container_name in got_container_resgroup_names:
```
jellonek: This if is not needed at all as `got_container_resgroup_names.get` will return `None` for non existing name. It would be better to assert that expected_resgroup_name should be set to some value, so `assert expected_resgroup_name is not None` or `assert isinstance(expected_regroup_name, str)` but the second case should be caught by later equality assert in this case.

owca/containers.py
iwankgb: We (@felidadae and myself) have just had a chat about this issue and we belive that there are two options:
* create two independent functions: `_find_new_tasks()` and `_find_dead_containers()`
* leave it as it is.

It's up to you, sir.

owca/platforms.py
jellonek: didn't know that, thx :D

owca/platforms.py
iwankgb: @ppalucki if you make it private then it's fine with me.

tests/test_runner.py
```diff
@@ -29,14 +29,15 @@
...
-        return Container(cgroup_path, rdt_enabled=False)
+        return Container(cgroup_path, rdt_enabled=False, platform_cpus=1)
 
 
 def metric(name, labels=None):
     """Helper method to create metric with default values. Value is ignored during tests."""
     return Metric(name=name, value=1234, labels=labels or {})
 
 
+@pytest.mark.skip('will be rewritten in next PR')
```
jellonek: btw. imo that's simply wrong. if something is changed in this PR - it should be fixed in the same PR.
if PR will grow because of that beyond some internal rules - fine - it's only about the test and breaking them because of change limit seems terribly wrong.
please do not repeat my mistakes, i mean - do not introduce something what is changing API and does not update the tests for the same api ;) (i made such mistake too many times and in long time it was (as this would if it will be leaved) too painful in the result...).

owca/platforms.py
iwankgb: @ppalucki SSG team does not mind using `Platform.rdt` property to represent an object storing all RDT-related platform information. Proceed with changed, please.

tests/test_runner.py
jellonek: I heard from @iwankgb that I've missed that this PR is against not a master branch and whole branch with provided tests is planned to be merged after bunch of other PRs - this means that my above comment was incorrect, so... Sorry for not needed noise ;)

owca/containers.py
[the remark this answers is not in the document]
iwankgb: It is.

tests/test_runner.py
ppalucki: It is very good advice, but we haven't found any better way of merging huge PR and disabling tests that we be fixed in future.

owca/containers.py
ppalucki: lets leave it as is

tests/test_containers.py
ppalucki: I did as you said, but then I have to explicitly pass information that this case expectes None,or I have an exception - that None doesn't equal t2. It just doesn't work this way. Please read my comment [here](https://github.com/intel/owca/pull/54#discussion_r260785732)

owca/platforms.py
ppalucki: @felidadae very good advice - lets do this ! (we got confirmaton) that this is acceptable for other team (see this [comment](https://github.com/intel/owca/pull/54#discussion_r261139584))

owca/platforms.py
ppalucki: done - it is private now

owca/platforms.py
ppalucki: @iwankgb I asked @felidadae to help me with that - to not block this PR further.

owca/platforms.py
ppalucki: is it blocking ?

owca/platforms.py
felidadae: fixed

owca/platforms.py
iwankgb: It's been resolved as far as I can see.

PR #57, opened 2019-02-27
Kubernetes file containing KubernetesNode and KubernetesTask

mmucek95:
Pull request to follow: https://github.com/ppalucki/owca/pull/6


owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+                    for unit in CPU_UNIT:
+                        if resource_value.endswith(unit):
+                            value = float(
+                                resource_value.split(unit)[0]) * CPU_UNIT[unit]
+                            break
+
+                    if not value:
+                        value = resource_value
+
+                if resource_type + '_' + resource_name in resources:
```
iwankgb: Extract `resource_type + '_' + resource_name` to helper variable, eg:
```python
resource_key = resource_type + '_' + resource_name
```
as it is used 3 times in 4 lines of code ;)

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+        return ('/kubepods/'
+                '{qos}/'
+                'pod{pod_id}/'
+                '{container_subdirectory}'.format(
+                            qos=qos.lower(),
+                            pod_id=pod_id,
+                            container_subdirectory=container_subdirectory))
+
+
+def _find_resources_for_pod(containers_spec):
```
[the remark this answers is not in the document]
iwankgb: Consider renaming function to `_calculate_pod_resources()` as you do not try to find anything but you simply multiply and add numbers :)

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+        for resource_type in RESOURCE_TYPES:
+            resource_exists = bool(container_resources.get(resource_type))
+            if not resource_exists:
+                continue
+
+            for resource_name, resource_value in \
+                    container_resources.get(resource_type).items():
+                value = None
+                if resource_name == 'memory':
+                    for unit in MEMORY_UNITS:
```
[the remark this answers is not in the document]
iwankgb: It might be reasonable to catch `KeyError` and log information about not recognized resource names.

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+
+                if resource_type + '_' + resource_name in resources:
+                    resources[resource_type + '_' + resource_name] += float(value)
+                else:
+                    resources[resource_type + '_' + resource_name] = float(value)
+
+    return resources
+
+
+def sanitize_label(label_key):
```
iwankgb: Make function private, please.

owca/testing.py
```diff
@@ -103,15 +103,16 @@ def anomaly(contended_task_id: TaskId, contending_task_ids: List[TaskId],
...
     """Helper method to create task with default values."""
     prefix = cgroup_path.replace('/', '')
     return Task(
         cgroup_path=cgroup_path,
         name=prefix + '_tasks_name',
         task_id=prefix + '_task_id',
         labels=labels or dict(),
-        resources=resources or dict()
+        resources=resources or dict(),
+        subcgroups_paths=subcgroups_paths if subcgroups_paths is not None else []
```
iwankgb: Can't it be:
```suggestion
        subcgroups_paths=subcgroups_paths or []
```
as it is in the line above?

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+    task = tasks[0]
+    assert task == KubernetesTask(name='test',
+                                  task_id='4d6a81df-3448-11e9-8e1d-246e96663c22',
+                                  qos='burstable',
+                                  labels={'exampleKey': 'value'},
+                                  resources={'requests_cpu': 0.25,
+                                             'requests_memory': 64*1024**2},
+                                  cgroup_path='/kubepods/burstable/pod4d6a81df-3448-11e9-8e'
+                                              '1d-246e96663c22/',
+                                  subcgroups_paths=['/kubepods/burstable/pod4d'
```
iwankgb: Is this a list with 5 elements or a string concatenated from 5 strings?

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+
+    task = tasks[1]
+    assert task == KubernetesTask(name='test2',
+                                  task_id='567975a0-3448-11e9-8e1d-246e96663c22',
+                                  qos='besteffort',
+                                  labels={},
+                                  resources={},
+                                  cgroup_path='/kubepods/besteffort/pod567975a0-3448-11e9-8e'
+                                              '1d-246e96663c22/',
+                                  subcgroups_paths=['/kubepods/besteffort/pod5'
```
iwankgb: Is this a list with 5 elements or a string concatenated from 5 strings?

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+
+
+@patch('requests.get', return_value=create_json_fixture_mock('kubernetes_get_state', 200))
+def test_get_tasks(get_mock):
+    node = KubernetesNode()
+    tasks = node.get_tasks()
+    assert len(tasks) == 2
+
+    task = tasks[0]
+    assert task == KubernetesTask(name='test',
```
iwankgb: Assert directly against `task[0]`, please.

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+                                  cgroup_path='/kubepods/burstable/pod4d6a81df-3448-11e9-8e'
+                                              '1d-246e96663c22/',
+                                  subcgroups_paths=['/kubepods/burstable/pod4d'
+                                                    '6a81df-3448-11e9-8e1d-246'
+                                                    'e96663c22/eb9c378219b6a4e'
+                                                    'fc034ea8898b19faa0e27c7b2'
+                                                    '0b8eb254fda361cceacf8e90'])
+
+    task = tasks[1]
+    assert task == KubernetesTask(name='test2',
```
iwankgb: Assert directly against `task[1]`, please.

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+                       'volumeMounts': [{'mountPath':
+                                         '/var/run/secrets/kubernetes.io/'
+                                         'serviceaccount',
+                                         'name': 'default-token-ktvvz',
+                                         'readOnly': True}]}]
+    assert {'requests_cpu': 0.35, 'requests_memory':
+            67108864 + 32 * 1024 ** 2} == _find_resources_for_pod(container_spec)
+
+
+def test_find_cgroup_path_for_pod_systemd():
```
iwankgb: What about cgroupfs case? :)

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+                                  cgroup_path='/kubepods/besteffort/pod567975a0-3448-11e9-8e'
+                                              '1d-246e96663c22/',
+                                  subcgroups_paths=['/kubepods/besteffort/pod5'
+                                                    '67975a0-3448-11e9-8e1d-24'
+                                                    '6e96663c22/e90bbbb3b060ba'
+                                                    'a1d354cd9b26f353d66fbb08d'
+                                                    '785abd32f4f6ec52ac843a2e7'])
+
+
+@patch('requests.get', return_value=create_json_fixture_mock('kubernetes_get_state_not_ready', 200))
```
iwankgb: Does this file consists of single pod?

tests/test_kubernetes.py
```diff
@@ -0,0 +1,144 @@
...
+def test_get_tasks_not_all_ready(get_mock):
+    node = KubernetesNode()
+    tasks = node.get_tasks()
+    assert len(tasks) == 0
+
+
+def test_find_resources_empty():
```
iwankgb: What do you think about passing spec consisting of resource only? Other fields are ignored anyway.

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+            # Ignore Kubernetes internal pods.
+            if pod.get('metadata').get('namespace') not in self.monitored_namespaces:
+                continue
+
+            # Read into variables essential information about pod.
+            pod_id = pod.get('metadata').get('uid')
+            pod_name = pod.get('metadata').get('name')
+            qos = pod.get('status').get('qosClass')
+            if pod.get('metadata').get('labels'):
+                labels = {sanitize_label(key): value
```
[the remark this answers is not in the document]
felidadae: to replace '.' '-' - otherwise:
def is_convertable_to_prometheus_exposition_format(metrics: List[Metric]) -> (bool, str):
will ignore the label
_METRIC_LABEL_NAME_RE = re.compile(r'^[a-zA-Z_][a-zA-Z0-9_]*$')
owca/storage.py:87

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+                      .format(pod_id, pod_name))
+
+            container_spec = pod.get('spec').get('containers')
+            tasks.append(KubernetesTask(name=pod_name, task_id=pod_id, qos=qos.lower(),
+                                        labels=labels,
+                                        resources=_find_resources_for_pod(container_spec),
+                                        cgroup_path=_find_cgroup_path_for_pod(self.cgroup_driver,
+                                                                              qos, pod_id),
+                                        subcgroups_paths=containers_cgroups))
+            self._log_found_tasks(tasks)
```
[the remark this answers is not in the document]
felidadae: ohh that was not on purpose :) i will move it just aboce return statement.

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+
+
+@dataclass
+class KubernetesNode(Node):
+    # We need to know cgroup driver used to properly construct cgroup paths for pods.
+    #   Reference in source code for kubernetes version stable 3.13:
+    #   https://github.com/kubernetes/kubernetes/blob/v1.13.3/pkg/kubelet/cm/cgroup_manager_linux.go#L207
+    #
+    # Use str as type (instead of CgroupDriverType) to
+    # simplify creation in the YAML config file.
```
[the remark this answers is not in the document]
felidadae: The configuration file would need to look like:

`
  node: !KubernetesNode
   # ... some fields
    cgroup_driver: !CgroupDriverType
       value: "systemd"
`
Apart from this I would need to register type CgroupDriverType in 
owca/component.py:
def register_components

Then I could remove assert from __post_init__ owca/kubernetes.py:71.py

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+                            pod_id=pod_id,
+                            container_subdirectory=container_subdirectory))
+
+
+def _find_resources_for_pod(containers_spec):
+    resources = dict()
+
+    MEMORY_UNITS = {'Ki': 1024, 'Mi': 1024**2, 'Gi': 1024**3, }
+    CPU_UNIT = {'m': 0.001}
+    RESOURCE_TYPES = ['requests', 'limits']
```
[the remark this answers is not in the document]
felidadae: They are used only in this function. Still?

owca/testing.py
felidadae: Yes, we should make it like that.

tests/test_kubernetes.py
felidadae: string concatenated from 5 strings inside a list

tests/test_kubernetes.py
felidadae: We test it in test_get_tasks - we should add to 32 line directly in the constructor that cgroup_driver="cgroupfs".
Still of course it may be useful to have only this one function tested in isolation. Should be add it (test_find_cgroup_path_for_pod_cgroupfs)?

tests/test_kubernetes.py
felidadae: @mmucek95 ?

tests/test_kubernetes.py
felidadae: I think it good idea to pass to find_resources_for_pod list of resources only to simplify tests.
So we would call
_find_resources_for_pod(pods_resources)
_find_resources_for_pod([container_spec.get('resources') for container_spec in container_specs])

Then we could just do what you mentioned

tests/test_kubernetes.py
mmucek95: There are also pods created by kubernetes but they are ignored. There is only one pod outside of kube-system namespace.

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # List of namespaces to monitor pods in.
+    monitored_namespaces: List[str] = field(default_factory=lambda: ["default"])
+
+    def __post_init__(self):
+        assert self.cgroup_driver in (CgroupDriverType.SYSTEMD, CgroupDriverType.CGROUPFS)
+
+    def _request_kubetelet(self):
+        METHOD = 'GET_STATE'
+        pods_path = '/pods'
```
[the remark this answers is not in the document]
felidadae: yes we use get:
`r = requests.get(full_url, json=dict(type=METHOD),`
`requests.get`
I think now that the best solution would be to remove that variable and just write it directly in 
`r = requests.get(full_url, json=dict(type="GET_STATE"),`
What You think?

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+    client_cert: str = None
+
+    # List of namespaces to monitor pods in.
+    monitored_namespaces: List[str] = field(default_factory=lambda: ["default"])
+
+    def __post_init__(self):
+        assert self.cgroup_driver in (CgroupDriverType.SYSTEMD, CgroupDriverType.CGROUPFS)
+
+    def _request_kubetelet(self):
+        METHOD = 'GET_STATE'
```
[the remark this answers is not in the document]
felidadae: yes we use get:
`r = requests.get(full_url, json=dict(type=METHOD),`
`requests.get`
I think now that the best solution would be to remove that variable and just write it directly in 
`r = requests.get(full_url, json=dict(type="GET_STATE"),`
What You think?

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from enum import Enum
+from dataclasses import dataclass, field
+from typing import List
+import logging
+import requests
+import urllib.parse
```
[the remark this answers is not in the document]
felidadae: build in package we can import it like that

owca/kubernetes.py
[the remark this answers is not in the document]
mmucek95: There are only two possible resource names so I don't really see why we should catch KeyError.

https://kubernetes.io/docs/concepts/configuration/manage-compute-resources-container/#resource-types

tests/test_kubernetes.py
iwankgb: I believe that it should be tested in isolation too or this tests should be removed completely and both scenarios should be covered in `test_get_tasks()`.

tests/test_kubernetes.py
```diff
@@ -0,0 +1,113 @@
...
+    node = KubernetesNode()
+    tasks = node.get_tasks()
+    assert len(tasks) == 2
+    assert tasks[0] == KubernetesTask(name='test',
+                                  task_id='4d6a81df-3448-11e9-8e1d-246e96663c22',
+                                  qos='burstable',
+                                  labels={'exampleKey': 'value'},
+                                  resources={'requests_cpu': 0.25,
+                                             'requests_memory': 64*1024**2},
+                                  cgroup_path='/kubepods/burstable/pod4d6a81df-3448-11e9-8e'
```
iwankgb: `cgroup_path` ends with `/` while elements in `subcgroups_paths` do not. This is inconsistent and should be unified.

owca/kubernetes.py
```diff
@@ -0,0 +1,220 @@
...
+
+    units = {'memory': MEMORY_UNITS, 'cpu': CPU_UNITS}
+
+    for container in containers_spec:
+        container_resources = container.get('resources')
+        if not container_resources:
+            continue
+
+        for resource_type in RESOURCE_TYPES:
+            resource_exists = bool(container_resources.get(resource_type))
```
iwankgb:
```suggestion
```

owca/kubernetes.py
```diff
@@ -0,0 +1,220 @@
...
+            if not resource_exists:
```
iwankgb:
```suggestion
            if resource_type not in container_resources:
```

owca/kubernetes.py
```diff
@@ -0,0 +1,220 @@
...
+
+            for resource_name, resource_value in \
+                    container_resources.get(resource_type).items():
+                value = None
+                for unit, multiplier in units.get(resource_name).items():
+                    if resource_value.endswith(unit):
+                        value = float(resource_value.split(unit)[0]) * multiplier
+                        break
+
+                if not value:
```
iwankgb: Instead of using this `if` statement you can do:
```python
value = resource_value
```
instead of 
```python
value = None
```
inline 194 :)

tests/test_kubernetes.py
felidadae: After discussion:
function input argument stays as it is. But in test we dont need to prepare the whole dictionary, just leave key 'resource'

owca/kubernetes.py
```diff
@@ -0,0 +1,220 @@
...
+                resource_key = resource_type + '_' + resource_name
+                if resource_key in resources:
+                    resources[resource_key] += float(value)
+                else:
+                    resources[resource_key] = float(value)
+
+    return resources
+
+
+def _sanitize_label(label_key):
```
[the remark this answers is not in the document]
felidadae: In case of mesos, the function also removes unwanted prefixes
`def sanitize_mesos_label(label_key):
    """Removes unwanted prefixes from Aurora & Mesos e.g. 'org.apache.aurora'
    and replaces invalid characters like "." with underscore.
    """
    # Drop unwanted prefixes
    for unwanted_prefix in MESOS_LABELS_PREFIXES_TO_DROP:
        if label_key.startswith(unwanted_prefix):
            label_key = label_key.replace(unwanted_prefix, '')
`

owca/kubernetes.py
```diff
@@ -0,0 +1,229 @@
...
+
+    # By default use localhost, however kubelet may not listen on it.
+    kubernetes_agent_enpoint: str = 'https://127.0.0.1:10250'
+
+    # Key and certificate to access kubelet API.
+    client_private_key: str = None
+    client_cert: str = None
+
+    # List of namespaces to monitor pods in.
+    monitored_namespaces: List[str] = field(default_factory=lambda: ["default"])
```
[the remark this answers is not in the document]
felidadae: https://docs.python.org/3/library/dataclasses.html#dataclasses.field
field(default_factory=list)
as default value should be mutable object List we need to use field function to set default_factory

PR #59, opened 2019-03-04
Kubernetes[2]: containers, add ContainerSet and needed changes to ContainerManager.

felidadae:
Full Kubernetes detection.


owca/resctrl.py
```diff
@@ -235,8 +237,14 @@ def _write_schemata_line(value, schemata_file):
                 log.error('Cannot set rdt allocation: {}'.format(e))
 
         with open(os.path.join(self.fullpath, SCHEMATA), 'bw') as schemata_file:
```
iwankgb: If file is opened with `w` mode then if will be truncated before writing. Is this what we want? If memory bandwidth and LLC were allocated in previous iteration and now *only* LLC information is passed then memory bandwidth allocation will be lost.

owca/runners/allocation.py
```diff
@@ -69,9 +69,9 @@ def create(tasks_allocations: TasksAllocations, containers, platform):
         def rdt_allocation_value_constructor(rdt_allocation: RDTAllocation, container,
```
iwankgb: Type hint for `container` is missing; shouldn't it be `ContainerInterface`?

owca/runners/allocation.py
```diff
@@ -69,9 +69,9 @@ def create(tasks_allocations: TasksAllocations, containers, platform):
...
                                              common_labels):
             return RDTAllocationValue(
-                container.container_name,
+                container.get_name(),
                 rdt_allocation,
-                container.resgroup,
+                container.get_resgroup(),
                 container.cgroup.get_pids,
```
iwankgb: This line won't work for `ContainerInterface`; shouldn't it be: `container.get_pids()`?

owca/testing.py
```diff
@@ -116,15 +116,50 @@ def task(cgroup_path, labels=None, resources=None, subcgroups_paths=None):
...
-            cgroup_path,
-            rdt_enabled=False, platform_cpus=1,
-            allocation_configuration=AllocationConfiguration() if with_config else None,
-            resgroup=ResGroup(name=resgroup_name) if resgroup_name is not None else None
-        )
+def container(cgroup_path, subcgroups_paths=[], resgroup_name=None,
+              with_config=False, should_patch=True):
+    """Helper method to create container optionally with patched subsystems."""
+    def unpatched():
+        if len(subcgroups_paths):
```
iwankgb: Do we really need this if statement? `containerset()` function is defined just below this one. It makes me think that this one should not have been modified that much.

owca/cgroups_allocations.py
```diff
@@ -37,9 +37,10 @@ def perform_allocations(self):
 
 class SharesAllocationValue(BoxedNumeric):
 
-    def __init__(self, normalized_shares: float, container: Container, common_labels: dict):
+    def __init__(self, normalized_shares: float, container: ContainerInterface,
+                 common_labels: dict):
```
iwankgb:
```suggestion
                 common_labels: Dict[str, str]):
```

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+    @abstractmethod
+    def get_cgroup(self) -> Cgroup:
+        ...
+
+    @abstractmethod
+    def get_cgroup_path(self) -> Cgroup:
+        ...
+
+    @abstractmethod
+    def get_pids(self) -> List[str]:
```
iwankgb: Aren't pids integers?

tests/runners/test_runners_detection.py
```diff
@@ -59,49 +72,35 @@ def test_detection_runner_containers_state(*mocks):
...
-        node=node_mock,
-        metrics_storage=metrics_storage,
-        anomalies_storage=anomalies_storage,
-        detector=detector_mock,
-        rdt_enabled=False,
-        extra_labels=extra_labels,
-    )
+    contended_task_id = 'task1'
+    contending_task_ids = ['task2']
+    anomaly_ = anomaly(contended_task_id=contended_task_id, contending_task_ids=contending_task_ids,
```
iwankgb: Why "_"? @felidadae, this is you, isn't it? ;)

tests/runners/test_runners_detection.py
```diff
@@ -117,37 +116,39 @@ def test_detection_runner_containers_state(*mocks):
...
+        {'t1_task_id': {'cpu_usage': CPU_USAGE * (len(subcgroups) if subcgroups else 1)}},
+        {'t1_task_id': {'cpus': CPU_COUNT}},
+        {'t1_task_id': task_labels_sanitized_with_task_id})
 
-    # assert expected state (new container based on first task /t1)
-    assert (runner.containers_manager.containers ==
-            {task('/t1', resources=dict(cpus=8.), labels=task_labels): container('/t1')})
+    # Assert expected state of runner.containers_manager (new container based on first task /t1).
+    assert len(runner.containers_manager.containers) == 1
+    task_, container_ = list(runner.containers_manager.containers.items())[0]
```
iwankgb: Stop adding trailing underscore to variable names. If it means something then I don't get it.

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
-                               PerfCoutners_mock, clean_mock, ResGroup_mock,
-                               tasks, existing_containers, mon_groups_relation,
-                               expected_running_containers):
-
-    containers_manager = ContainerManager(
-        rdt_enabled=True,
-        rdt_mb_control_enabled=False,
-        platform_cpus=1,
-        allocation_configuration=AllocationConfiguration(),
+    'tasks_, existing_containers_, mon_groups_relation, expected_running_containers_',
```
iwankgb: Why have underscores been added? Does it have any special meaning?

iwankgb: What was wrong with tasks and containers being created in `parametrize` call?

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
+def assert_equal_containers(containers_a: List[ContainerInterface],
+                            containers_b: List[ContainerInterface]):
+    """Compares two list of containers. IMPORTANT: two containers are HERE
+       treated as equal if they have the same cgroup_path set.
+
+       One special assumption is made about private attribute of
+       ContainerSet: that field _subcontainers exists with type
+       dict[_, Container]."""
+    assert len(containers_a) == len(containers_b)
+    for a, b in zip(containers_a, containers_b):
```
iwankgb: There is an additional assumption here: order of items in `containers_a` and `containers_b` is identical.

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
 from owca.allocators import AllocationConfiguration
 
 
+def assert_equal_containers(containers_a: List[ContainerInterface],
+                            containers_b: List[ContainerInterface]):
+    """Compares two list of containers. IMPORTANT: two containers are HERE
+       treated as equal if they have the same cgroup_path set.
+
+       One special assumption is made about private attribute of
+       ContainerSet: that field _subcontainers exists with type
```
iwankgb: Consider declaring `__eq__` on `Container` and `ContainerSet`.

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
-         [], []),  # no actions
-        ([task('/t1'), task('/t2')], [container('/t1'), ],  # another task arrived,
-         [task('/t2')], []),  # let's create another container,
-        ([task('/t1'), task('/t2')], [container('/t1'), container('/t2')],  # 2on2 converged
-         [], []),  # nothing to do,
-        ([task('/t2')], [container('/t1'), container('/t2')],  # first task just disappeared
-         [], [container('/t1')]),  # remove the first container
-        # some other cases
-        ([task('/t1'), task('/t2')], [],  # the new task, just appeared
+    'discovered_tasks, containers, '
```
iwankgb: Has anything really changed here? I am a bit puzzled to be honest.

tests/runners/test_runners_detection.py
```diff
@@ -11,41 +11,54 @@
...
 @patch('owca.runners.base.are_privileges_sufficient', return_value=True)
 @patch('owca.containers.ResGroup')
 @patch('owca.containers.PerfCounters')
 @patch('owca.platforms.collect_topology_information', return_value=(1, 1, 1))
-@patch('owca.containers.Cgroup.get_measurements', return_value=dict(cpu_usage=23))
-@patch('time.time', return_value=1234567890.123)
-def test_detection_runner_containers_state(*mocks):
+@patch('owca.containers.Cgroup.get_measurements', return_value=dict(cpu_usage=CPU_USAGE))
+@patch('time.time', return_value=TIME_TICK)
+@pytest.mark.parametrize('subcgroups', ([], ['/t1/c1'], ['/t1/c1', '/t1/c2']))
```
iwankgb: Make it first argument and ignore all others.

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
+    containerset_ = containerset('/t1', subcgroups_paths, should_patch=False)
+
+    resgroup_mock = Mock(spec=ResGroup, get_measurements=Mock(return_value={'foo': 3}))
+    containerset_.set_resgroup(resgroup=resgroup_mock)
+
+    # Call the main function.
+    measurements = containerset_.get_measurements()
+
+    resgroup_mock.get_measurements.assert_called_once()
+    assert {'foo': 3, 'cgroup_metric__1': ANY_METRIC_VALUE*N,
```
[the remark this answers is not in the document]
felidadae: That was on purpose. In this way I do not place in code some magic numbers, where reader must guess how it was calculated, but show them the formula. I believe that this is more readibly - once i read @ppalucki tests without it I just were confused.

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
+            'perf_event_metric__1': ANY_METRIC_VALUE*N} == measurements
```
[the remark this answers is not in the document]
felidadae: That was on purpose. In this way I do not place in code some magic numbers, where reader must guess how it was calculated, but show them the formula. I believe that this is more readibly - once i read @ppalucki tests without it I just were confused.

owca/containers.py
felidadae: cgroup.get_pids returns string, as it reads a file and return non empty lines.

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+    @abstractmethod
+    def set_resgroup(self, resgroup: ResGroup):
+        ...
+
+    @abstractmethod
+    def get_resgroup(self):
+        ...
+
+    @abstractmethod
+    def get_name(self):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
-    rdt_enabled: bool = True
-    rdt_mb_control_enabled: bool = False
-    container_name: str = None  # defaults to faltten value of provided cgroup_path
+class ContainerInterface(ABC):
+    @abstractmethod
+    def set_resgroup(self, resgroup: ResGroup):
+        ...
+
+    @abstractmethod
+    def get_resgroup(self):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/cgroups_allocations.py
felidadae: fixed

docs/detection.rst
```diff
@@ -19,6 +19,13 @@ Check `Prometheus metrics types`_ for further reference
 .. _`Prometheus metrics types`: https://prometheus.io/docs/concepts/metric_types
 
 
+Supported orchestration systems
+-----------
+We support both Mesos and Kubernetes (Kubernetes is currently supported only in the detect mode).
+Use `config kubernetes`_ as a reference configuration file.
+
+.. _`config kubernetes`: configs/kubernetes_example.yaml
```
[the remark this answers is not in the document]
felidadae: fixed. just referencing configs directory as a whole

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+    def cleanup(self) -> None:
+        ...
+
+    @abstractmethod
+    def get_allocations(self) -> TaskAllocations:
+        ...
+
+
+class ContainerSet(ContainerInterface):
+    def __init__(self,
```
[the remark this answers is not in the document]
felidadae: To be consistent I would also need to change Container class. I dont think it is scope of this PR to do this.

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+        self._resgroup = resgroup
+
+        # Create Cgroup object representing itself.
+        self._cgroup = Cgroup(
+            cgroup_path=self._cgroup_path,
+            platform_cpus=platform_cpus,
+            allocation_configuration=allocation_configuration)
+
+        # Create Cgroup objects for children.
+        self._subcontainers: Dict[str, Container] = {}
```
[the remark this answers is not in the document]
felidadae: I do know that i keep objects of class Container. The typing is here just to improve readibility. It stress the fact, that subcontainers cannot be of type ContainerSet, but only of type Container.

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+                # Managament of RDT is done only in ContainerSet class.
+                rdt_enabled=False,
+                rdt_mb_control_enabled=False,
+                platform_cpus=platform_cpus,
+                allocation_configuration=allocation_configuration)
+
+    def set_resgroup(self, resgroup: ResGroup):
+        self._resgroup = resgroup
+
+    def get_resgroup(self):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+            self._resgroup.add_pids(pids=self.get_pids(), mongroup_name=self._name)
+
+    def get_measurements(self) -> Measurements:
+        measurements = dict()
+        try:
+            # Merge cgroup and perf_counters measurements.
+            measurements_list: List[Measurements] = []
+            for container in self._subcontainers.values():
+                m = container.cgroup.get_measurements()
+                m.update(container.perf_counters.get_measurements())
```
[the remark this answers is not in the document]
felidadae: Fixed. It was done this in that way, as previously I did not set rdt_enabled to false for self._subcontainers.

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+                 rdt_enabled: bool = True, rdt_mb_control_enabled: bool = False):
+        self._cgroup_path = cgroup_path
+        self._name = _sanitize_cgroup_path(self._cgroup_path)
+        self._allocation_configuration = allocation_configuration
+        self._rdt_enabled = rdt_enabled
+        self._rdt_mb_control_enabled = rdt_mb_control_enabled
+        self._resgroup = resgroup
 
-    def __post_init__(self):
         self.cgroup = Cgroup(
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
-            platform_cpus=self.platform_cpus,
-            allocation_configuration=self.allocation_configuration,
-        )
-        self.container_name = (self.container_name or
-                               _sanitize_cgroup_path(self.cgroup_path))
-        self.perf_counters = PerfCounters(self.cgroup_path, event_names=DEFAULT_EVENTS)
+            cgroup_path=self._cgroup_path,
+            platform_cpus=platform_cpus,
+            allocation_configuration=allocation_configuration)
+        self.perf_counters = PerfCounters(self._cgroup_path, event_names=DEFAULT_EVENTS)
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -52,36 +52,194 @@ def _sanitize_cgroup_path(cgroup_path):
...
+
+    def set_resgroup(self, resgroup: ResGroup):
+        self._resgroup = resgroup
+
+    def get_cgroup(self):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -94,38 +252,38 @@ def get_measurements(self) -> Measurements:
...
         # In only detect mode, without allocation configuration return nothing.
-        if not self.allocation_configuration:
+        if not self._allocation_configuration:
             return {}
+
         allocations: TaskAllocations = dict()
-        allocations.update(self.cgroup.get_allocations())
-        if self.rdt_enabled:
-            allocations.update(self.resgroup.get_allocations())
+        allocations.update(self._cgroup.get_allocations())
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -178,30 +337,44 @@ def sync_containers_state(self, tasks) -> Dict[Task, Container]:
...
-                new_task.cgroup_path,
-                rdt_enabled=self.rdt_enabled,
-                rdt_mb_control_enabled=self.rdt_mb_control_enabled,
-                platform_cpus=self.platform_cpus,
-                allocation_configuration=self.allocation_configuration,
-            )
+            # Check whether the task groups multiple containers,
+            #   is so use ContainerSet class, otherwise Container class.
+            #   ContainerSet shares interface with Container.
+            if len(new_task.subcgroups_paths):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/metrics.py
```diff
@@ -110,3 +110,28 @@ def create_metric_with_metadata(name, value, labels=None):
 
 
 Measurements = Dict[MetricName, MetricValue]
+
+
+def sum_measurements(measurements_list: List[Measurements]) -> \
+        Tuple[Measurements, List[MetricName]]:
+    """Returns dictionary with metrics which are contained in all input measurements with value set
+       to arithmetic sum."""
+    sum_: Measurements = {}
```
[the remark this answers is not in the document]
felidadae: fixed. sum_ -> summed_metrics. sum is buildin function which sums an iterable.

owca/metrics.py
```diff
@@ -110,3 +110,28 @@ def create_metric_with_metadata(name, value, labels=None):
...
+
+    common_metrics = set()  # Intersect of set of names.
+    for measurements in measurements_list:
+        metrics_names = {metric_name for metric_name in measurements.keys()}
```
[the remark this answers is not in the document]
felidadae: set(measurements.keys())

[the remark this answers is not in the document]
felidadae: fixed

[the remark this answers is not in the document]
felidadae: Additionally, to difference between variables keeping metrics with those keeping metrics names I added appropriate suffix to their names:
summed_metrics
common_metrics_names

owca/resctrl.py
felidadae: There should not be any changes in this file in this PR. I will try to diagnose the reason for this.

owca/runners/allocation.py
felidadae: fixed

owca/runners/allocation.py
felidadae: fixed

tests/test_containers.py
felidadae: I see such disadvantages of adding __eq__:
* only needed in unit tests, not in real code
* not clear when two containers are equal (like in assert_equal_containers, or should also I check other fields in Container and ContainerSet class)
* and another associated with the previous one: as it is not clear it may lead to some subtle bugs, when sb expected function to behave differently

tests/test_containers.py
felidadae: I added that to docstring:
`Compares two lists (order does have importance) of containers.`

tests/test_containers.py
felidadae: Just to IMHO improve readibility by putting new line in the same way as parametrs below are written.
Example:

'discovered_tasks, containers, ' >>NEW_LINE<<
'expected_new_tasks, expected_containers_to_delete'

([task('/t1')], [], >>NEW_LINE<<
[task('/t1')], []),

tests/test_containers.py
felidadae: We need to pass parameter from parametrize above
`@pytest.mark.parametrize('subsgroups', ([], ['/t1/c1', '/t1/c2']))`
to function `container`. We cannot do it differently than inside body of `test_sync_containers_state`
That is the reason for keeping underscores. I could name it differently, but it's not simple to make better names and whats more important it corresponds 1:1 to arguments needed to create real objects.
E.g. existing_containers_ is of type Dict[CgroupPath, CgroupPath] and real objects are Dict[Task, ContainerInterface].
Should I make out different names or add comment explaining why I use underscore?

tests/test_containers.py
```diff
@@ -12,101 +12,195 @@
...
+@patch('owca.containers.PerfCounters', spec=PerfCounters,
+       get_measurements=Mock(return_value={'perf_event_metric__1': ANY_METRIC_VALUE}))
+def test_containerset_get_measurements(*args):
+    """Check whether summing of metrics for children containers are done properly.
+       Note: because we are mocking here classes from which measurements are read,
+       to calculate the proper value of ContainerSet we just need to multiple that
+       single value by count of subcontainers (here defined as N)."""
+    N = 3  # 3 subcontainers are created.
+    subcgroups_paths = ['/t1/c1', '/t1/c2', '/t1/c3']
+    containerset_ = containerset('/t1', subcgroups_paths, should_patch=False)
```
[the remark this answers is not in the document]
felidadae: Here reason for underscore was that containerset is already used name for function.

tests/runners/test_runners_detection.py
felidadae: It cannot be changed by changing position of the usage of decorator `@pytest.mark.parametrize('subcgroups', ([], ['/t1/c1'], ['/t1/c1', '/t1/c2']))`. It seems that `parametrize` makes such an impact on decorated function that expects resulting arguments to be the last in order.

tests/runners/test_runners_detection.py
felidadae: fixed. it was used, as task and container are already used names -- they are names of functions

tests/test_metrics.py
```diff
@@ -41,3 +41,13 @@ def test_sanitize_labels(label_key, expected_label_key):
 def test_create_metrics(task_measurements, expected_metrics):
     got_metrics = create_metrics(task_measurements)
     assert expected_metrics == got_metrics
+
+
+@pytest.mark.parametrize('measurements_list,expected_sum,expected_ignored', (
+    ([{'m1': 8}, {'m1': 3}], {'m1': 11}, []),
+    ([{'m1': 8}, {'m1': 3, 'm2': 2}], {'m1': 11}, ['m2']),
+    ([{'m1': 8}, {'m1': 3}, {'m1': 3}], {'m1': 14}, []),
+    ([{'m1': 8, 'm2': 3}, {'m1': 3, 'm2': 7}, {'m1': 3}], {'m1': 14}, ['m2']),
```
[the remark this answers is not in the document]
felidadae: fixed

tests/test_metrics.py
```diff
@@ -41,3 +41,13 @@ def test_sanitize_labels(label_key, expected_label_key):
 def test_create_metrics(task_measurements, expected_metrics):
     got_metrics = create_metrics(task_measurements)
     assert expected_metrics == got_metrics
+
+
+@pytest.mark.parametrize('measurements_list,expected_sum,expected_ignored', (
+    ([{'m1': 8}, {'m1': 3}], {'m1': 11}, []),
+    ([{'m1': 8}, {'m1': 3, 'm2': 2}], {'m1': 11}, ['m2']),
+    ([{'m1': 8}, {'m1': 3}, {'m1': 3}], {'m1': 14}, []),
```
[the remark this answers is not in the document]
felidadae: fixed

owca/testing.py
felidadae: I removed containerset funciton, as it was not needed.

owca/containers.py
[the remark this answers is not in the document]
iwankgb: Why do you think that they can't be of `ContainerSet` type? Are there any interactions with `self._subcontainers` that access their properties directly?

owca/containers.py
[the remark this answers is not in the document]
iwankgb: Why? Is a constructor somehow part of an interface?

tests/test_metrics.py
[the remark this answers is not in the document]
iwankgb: @felidadae I don't think that you added such a case.

owca/resctrl.py
iwankgb: @felidadae have you found reason for such approach?

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+
+            summed_measurements, ignored_metrics = sum_measurements(measurements_list)
+            measurements.update(summed_measurements)
+            if ignored_metrics:
+                log.warning('ContainerSet.get_measuremenets: ignored metrics {} while summing up.'
+                            .format(ignored_metrics))
+
+            # Resgroup management is entirely done in this class.
+            if self._rdt_enabled:
+                measurements.update(self._resgroup.get_measurements(self._name))
```
iwankgb: The only scenario when `Resgroup.get_measurements()` can raise `FileNotFoundError` is when we try to call `get_measurements()` on an object that was previously cleaned up (by calling `Resgroup.remove()` function). If something like this happens then it's a perfect reason for terminating OWCA.

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+
+        except FileNotFoundError:
```
iwankgb: Taking my two exception-related comments above: remove the `try...except` block, please.

owca/containers.py
```diff
@@ -94,40 +250,61 @@ def get_measurements(self) -> Measurements:
...
-        if self.rdt_enabled:
-            self.resgroup.remove(self.container_name)
+        self._perf_counters.cleanup()
+        if self._rdt_enabled:
+            self._resgroup.remove(self._name)
 
     def get_allocations(self) -> TaskAllocations:
         # In only detect mode, without allocation configuration return nothing.
-        if not self.allocation_configuration:
+        if not self._allocation_configuration:
```
iwankgb: This condition should be removed from `Container`. There is an assert in `Cgroup.get_allocations()` that handles this case. Exception must not be caught.

owca/containers.py
```diff
@@ -94,40 +250,61 @@ def get_measurements(self) -> Measurements:
...
 class ContainerManager:
     """Responsible for synchronizing state between found orchestration software tasks,
-    their containers and resctrl system.
-    """
+    their containers and resctrl system. """
 
     def __init__(self, rdt_enabled: bool, rdt_mb_control_enabled: bool, platform_cpus: int,
                  allocation_configuration: Optional[AllocationConfiguration]):
-        self.containers: Dict[Task, Container] = {}
+        self.containers: Dict[Task, ContainerInterface] = {}
```
iwankgb: Make all the properties private, please. There is not reason to expose internals of this class.

owca/containers.py
```diff
@@ -94,40 +250,61 @@ def get_measurements(self) -> Measurements:
...
         self.rdt_enabled = rdt_enabled
         self.rdt_mb_control_enabled = rdt_mb_control_enabled
         self.platform_cpus = platform_cpus
         self.allocation_configuration = allocation_configuration
 
+    def _create_container(self, task: Task) -> ContainerInterface:
```
iwankgb: This function is nearly identical to `testing.container()`. Taking into consideration that they share substantial amount of code shouldn't we reuse the code?
Another problem is that we use copy-pasted code in tests: it won't really check anything as we discussed earlier today at `test_runner.py` meeting.

owca/kubernetes.py
```diff
@@ -139,33 +138,29 @@ def get_tasks(self) -> List[Task]:
...
-                            qos=qos.lower(),
-                            pod_id=pod_id,
-                            container_subdirectory=container_subdirectory))
+        return ('/kubepods.slice'
+                '/kubepods-{qos}.slice'
+                '/kubepods-{qos}-pod{pod_id}.slice'
+                '/{container_subdirectory}/'.format(
+                    qos=qos,
+                    pod_id=pod_id,
+                    container_subdirectory=container_subdirectory)).replace('//', '/')
```
iwankgb: Remove `replace()` call, please.

owca/kubernetes.py
```diff
@@ -139,33 +138,29 @@ def get_tasks(self) -> List[Task]:
...
-                '{container_subdirectory}'.format(
-                            qos=qos.lower(),
-                            pod_id=pod_id,
-                            container_subdirectory=container_subdirectory))
+            container_subdirectory = container_id
+
+        qos = qos if qos != "guaranteed" else ""
+        return ('/kubepods/{qos}/pod{pod_id}/{container_subdirectory}/'.format(
+                    qos=qos, pod_id=pod_id,
+                    container_subdirectory=container_subdirectory)).replace('//', '/')
```
iwankgb: Same here: remove `replace()` call, please (and add any `if`s that are needed).

owca/testing.py
```diff
@@ -89,7 +89,8 @@ def anomaly_metrics(contended_task_id: TaskId, contending_task_ids: List[TaskId]
                                     type='contention',
                                     contending_workload_instance=contending_workload_instances[
                                         task_id], workload_instance=contending_workload_instances[
-                                            contended_task_id]), type=MetricType.COUNTER)
+                                            contended_task_id]),
+                        type=MetricType.COUNTER)
```
iwankgb: Why have you decided to make this change? If it does coding standard is not violated then undo it, please.

tests/test_containers.py
iwankgb: Comment explaining this case would help a lot. I had no idea that this is how two `pytest.mark.parametrize()` calls work.

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+                               tasks_, existing_containers_,
+                               mon_groups_relation, expected_running_containers_):
+    """Test both Container and ContainerSet classes.
+        Note: existing_containers, expected_running_containers and got_containers are of type:
+        Dict[Task, ContainerInterface]."""
+
+    # Create Task and Container/ContainerSet objects from strings given in parametrize.
+    #   This is done to both test Container and ContainerSet classes (passing
+    #   subsgroups parameter into constructing function >>container<<.
+    tasks = [task(t, subcgroups_paths=subsgroups) for t in tasks_]
```
iwankgb: s/tasks/running_tasks/
s/tasks_/tasks/

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+    existing_containers = {task(t, subcgroups_paths=subsgroups): container(c, subsgroups)
```
iwankgb: s/existing_containers/known_containers/

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+                           for t, c in existing_containers_.items()}
```
iwankgb: s/existing_containers_/containers/

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+    expected_running_containers = {task(t, subcgroups_paths=subsgroups): container(c, subsgroups)
+                                   for t, c in expected_running_containers_.items()}
```
iwankgb: s/expected_running_containers_/running_containers/

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+    # expected_running_containers are equal.
+    assert len(got_containers) == len(expected_running_containers)
+    assert all([expected_task in got_containers
+                for expected_task in expected_running_containers.keys()])
+    # Get List[ContainerInterface] from Dict[Task, ContainerInterface]. Keep the same order
+    # of elements in both expected_containers_ and got_containers_.
+    got_containers_list = [got_containers[task] for task in expected_running_containers.keys()]
+    expected_containers_list = [expected_running_containers[task]
+                                for task in expected_running_containers.keys()]
+    assert_equal_containers(got_containers_list, expected_containers_list)
```
iwankgb: I don't think that you need to transform both dicts to lists and pass them to `assert_equal_containers()`. Wouldn't something like this make sense?
```python
for task in expected_running_containers.keys():
    assert expected_running_containers[task] == got_containers[task]
```
It might be necessary to extract logic responsible for comparing two containers from `assert_equal_containers()`.

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+    got_container_resgroup_names = {c.get_name():
+                                    c.get_resgroup().name for c in got_containers.values()}
     for expected_resgroup_name, container_names in mon_groups_relation.items():
         for container_name in container_names:
             if container_name in got_container_resgroup_names:
                 got_resgroup_name = got_container_resgroup_names.get(container_name)
                 assert got_resgroup_name == expected_resgroup_name
+
+
+ANY_METRIC_VALUE = 2
```
iwankgb: Shouldn't it be private?

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+
+
+@patch('owca.containers.Cgroup', spec=Cgroup,
+       get_measurements=Mock(return_value={'cgroup_metric__1': ANY_METRIC_VALUE}))
+@patch('owca.containers.PerfCounters', spec=PerfCounters,
+       get_measurements=Mock(return_value={'perf_event_metric__1': ANY_METRIC_VALUE}))
+def test_containerset_get_measurements(*args):
+    """Check whether summing of metrics for children containers are done properly.
```
iwankgb: There is separate test for `sum_measurements()`. What are you really testing here?

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+       Note: because we are mocking here classes from which measurements are read,
+       to calculate the proper value of ContainerSet we just need to multiple that
+       single value by count of subcontainers (here defined as N)."""
+    N = 3  # 3 subcontainers are created.
+    subcgroups_paths = ['/t1/c1', '/t1/c2', '/t1/c3']
+    containerset_ = container('/t1', subcgroups_paths, should_patch=False, rdt_enabled=True)
```
iwankgb: "_".

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+
+    # Call the main function.
+    measurements = containerset_.get_measurements()
+
+    resgroup_mock.get_measurements.assert_called_once()
+    assert {'foo': 3, 'cgroup_metric__1': ANY_METRIC_VALUE*N,
+            'perf_event_metric__1': ANY_METRIC_VALUE*N} == measurements
+
+
+def smart_get_pids():
```
iwankgb: This function must be private.

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+
+
+@patch('owca.containers.Cgroup')
+@patch('owca.containers.Container', spec=Container, get_pids=Mock(side_effect=smart_get_pids()))
+def test_containerset_get_pids(*args):
+    subcgroups_paths = ['/t1/c1', '/t1/c2', '/t1/c3']
+    containerset_ = container('/t1', subcgroups_paths)
+    # We expect 6 consecutive numbers starting from 0 - as there are 3 subcgroups paths
+    # for the container.
+    assert containerset_.get_pids() == [str(i) for i in range(6)]
```
iwankgb:
```suggestion
    assert containerset_.get_pids() == [1, 2, 3, 4, 5, 6]
```

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+
+
+def smart_get_pids():
+    # Note: here List[int] is used instead of >>int<<
+    #   to pass mutable object (Integers are inmutable in Python).
+    calls_count = [0]
+
+    def fun() -> List[str]:
+        """Returns list of two consecutive integers. On first call
+           return [0,1], on second [2,3], and so forth."""
```
iwankgb: There is no pid 0 and even in tests we should avoiding using it :)

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+    containerset_ = container('/t1', subcgroups_paths)
+    # We expect 6 consecutive numbers starting from 0 - as there are 3 subcgroups paths
+    # for the container.
+    assert containerset_.get_pids() == [str(i) for i in range(6)]
+
+
+@patch('owca.containers.ResGroup.get_allocations', return_value={})
+@patch('owca.containers.Cgroup.get_allocations', return_value={})
+@patch('owca.containers.Container', spec=Container)
+def test_get_allocations(*mock):
```
iwankgb: This test makes no sense, remove please.

tests/test_kubernetes.py
```diff
@@ -77,39 +78,49 @@ def test_calculate_resources_empty():
...
-                       {'cpu': '250m',
-                        'memory': '64Mi'}}}]
-    assert {'limits_cpu': 0.25, 'limits_memory': 64*1024**2,
-            'requests_cpu': 0.25, 'requests_memory':
-                64*1024**2} == _calculate_pod_resources(container_spec)
+        {'resources': {'limits':   {'cpu': '250m', 'memory': '64Mi'},
+                       'requests': {'cpu': '250m', 'memory': '64Mi'}}}
+    ]
+    assert {'limits_cpu': 0.25,
+            'limits_memory': 64*1024**2,
```
iwankgb: Shouldn't it be `float` as `requests_memory` is?

tests/test_kubernetes.py
```diff
@@ -77,39 +78,49 @@ def test_calculate_resources_empty():
...
-    cgroup_driver = 'cgroupfs'
-    qos = 'burstable'
-    pod_id = '12345-67890'
-    assert '/kubepods/burstable/' \
-           'pod12345-67890/' == \
-           _build_cgroup_path(cgroup_driver, qos, pod_id)
+            float(67108864 + 32 * 1024 ** 2)} == _calculate_pod_resources(container_spec)
+
+
+POD_ID = '12345-67890'
```
iwankgb: It must be private.

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+    @abstractmethod
+    def get_name(self) -> str:
+        """Returns cgroup_path cleaned."""
+
+    @abstractmethod
+    def get_cgroup(self) -> Cgroup:
+        ...
+
+    @abstractmethod
+    def get_cgroup_path(self) -> Cgroup:
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+            cgroup_path=self._cgroup_path,
+            platform_cpus=platform_cpus,
+            allocation_configuration=allocation_configuration)
+
+        # Create Cgroup objects for children.
+        self._subcontainers: Dict[str, Container] = {}
+        for cgroup_path in cgroup_paths:
+            self._subcontainers[cgroup_path] = Container(
+                cgroup_path=cgroup_path,
+                # Managament of RDT is done only in ContainerSet class.
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+
+    def get_cgroup(self) -> Cgroup:
+        return self._cgroup
+
+    def get_cgroup_path(self) -> str:
+        return self._cgroup_path
+
+    def get_pids(self) -> List[str]:
+        all_pids = []
+        for container in self._subcontainers.values():
```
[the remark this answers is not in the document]
felidadae: _subcontainers are of type
`self._subcontainers: Dict[str, Container] = {}`
so the answer is no.

owca/containers.py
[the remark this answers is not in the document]
felidadae: Decided that it will not be changed because of `least knowledge principle`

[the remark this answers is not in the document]
iwankgb: @felidadae convinced me calling principle of least knowledge.

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+        """Called every run iteration to keep pids of cgroup and resctrl in sync."""
+        if self._rdt_enabled:
+            self._resgroup.add_pids(pids=self.get_pids(), mongroup_name=self._name)
+
+    def get_measurements(self) -> Measurements:
+        measurements = dict()
+        try:
+            # Merge cgroup and perf_counters measurements. As we set rdt_enabled to False
+            #   for subcontainers, it will ignore rdt measurements.
+            measurements_list: List[Measurements] = [container.get_measurements()
```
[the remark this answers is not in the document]
felidadae: fixed

owca/containers.py
```diff
@@ -53,37 +53,193 @@ def _sanitize_cgroup_path(cgroup_path):
...
+                                                     for container in self._subcontainers.values()]
```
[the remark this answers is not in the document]
felidadae: _subcontainers are of type
self._subcontainers: Dict[str, Container] = {}
so i cannot change it as you suggested

owca/containers.py
felidadae: I added the try-except clause in commit:
Fix race condition: the mesos task killed during runner iteration loop. (#27)

So I think it needs to stay, but only for line:
`measurements.update(self._resgroup.get_measurements(self._name))`

owca/containers.py
felidadae: Above answered

owca/containers.py
felidadae: https://github.com/intel/owca/commit/16df03c0a74badd52fa39270980fc86fb8cda9f3#diff-953e43a7e6cc771e02e854fecb26e3b1R57

felidadae: tested on cluster: only cgroup.get_measurements get throw this exception.
So now only this part is inside try except block.

felidadae: resgroups are managed only by us, so we do not expect any exceptions.

owca/containers.py
felidadae: fixed

owca/containers.py
felidadae: That was not edited in this PR. I will pass this comment to allocation PR.

owca/kubernetes.py
```diff
@@ -139,33 +138,29 @@ def get_tasks(self) -> List[Task]:
...
-        return ('/kubepods/'
-                '{qos}/'
-                'pod{pod_id}'
-                '{container_subdirectory}'.format(
-                            qos=qos.lower(),
-                            pod_id=pod_id,
-                            container_subdirectory=container_subdirectory))
+            container_subdirectory = container_id
+
+        qos = qos if qos != "guaranteed" else ""
```
ppalucki:
```suggestion
        qos = qos if qos != 'guaranteed' else ''
```

owca/kubernetes.py
felidadae: fixed

owca/runners/allocation.py
```diff
@@ -66,13 +66,13 @@ def create(tasks_allocations: TasksAllocations, containers, platform):
         # Shared object to optimize schemata write and detect CLOSids exhaustion.
         rdt_groups = RDTGroups(closids_limit=platform.rdt_information.num_closids)
 
-        def rdt_allocation_value_constructor(rdt_allocation: RDTAllocation, container,
-                                             common_labels):
+        def rdt_allocation_value_constructor(rdt_allocation: RDTAllocation,
+                                             container: ContainerInterface, common_labels):
```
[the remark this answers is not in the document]
felidadae: fixed

owca/testing.py
felidadae: it's unreadeable

owca/testing.py
```diff
@@ -121,15 +122,33 @@ def task(cgroup_path, labels=None, resources=None, subcgroups_paths=None):
...
-    """Helper method to create container with patched subsystems."""
-    with patch('owca.containers.ResGroup'), patch('owca.containers.PerfCounters'):
-        return Container(
-            cgroup_path,
-            rdt_enabled=False, platform_cpus=1,
-            allocation_configuration=AllocationConfiguration() if with_config else None,
-            resgroup=ResGroup(name=resgroup_name) if resgroup_name is not None else None
-        )
+def container(cgroup_path, subcgroups_paths=[], resgroup_name=None,
+              with_config=False, should_patch=True, rdt_enabled=False):
```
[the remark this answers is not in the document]
felidadae: fixed

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
 
+def assert_equal_containers(containers_a: List[ContainerInterface],
+                            containers_b: List[ContainerInterface]):
+    """Compares two lists (order does have importance) of containers.
+       IMPORTANT: two containers are HERE
+       treated as equal if they have the same cgroup_path set.
+
+       One special assumption is made about private attribute of
+       ContainerSet: that field _subcontainers exists with type
+       dict[_, Container]."""
```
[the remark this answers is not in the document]
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+        (['/t1'], {'/t1': '/t1', '/t2': '/t2'}, {'t2': ['t2']},
+         {'/t1': '/t1'}),
+        # 5) Two task (of two) dissapeared.
+        ([], {'/t1': '/t1', '/t2': '/t2'}, {},
+         {}),
     )
+)
+def test_sync_containers_state(get_pids_mock, sync_mock, perf_counters_mock,
+                               add_pids_mock, clean_taskless_groups_mock,
+                               subsgroups,
```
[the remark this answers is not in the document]
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: this are strings, not ints

tests/test_containers.py
```diff
@@ -12,101 +12,204 @@
...
+        calls_count[0] += 2
+        return [str(calls_count[0] - 2), str(calls_count[0] - 1)]
+    return fun
+
+
+@patch('owca.containers.Cgroup')
+@patch('owca.containers.Container', spec=Container, get_pids=Mock(side_effect=smart_get_pids()))
+def test_containerset_get_pids(*args):
+    subcgroups_paths = ['/t1/c1', '/t1/c2', '/t1/c3']
+    containerset_ = container('/t1', subcgroups_paths)
```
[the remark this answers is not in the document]
felidadae: it is ContainerSet as len(subcgroups_paths) == 3

tests/test_containers.py
felidadae: It has sense, it test if get_allocations do not use undefined variables or functions - before I use there unexisting variable self.cgroup instead of self._cgroup. It has little sense imho but it was suggested in first iteration of this PR review process.

tests/test_kubernetes.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed by 
`def assert_equal_containers(container_a: ContainerInterface,
                            container_b: ContainerInterface):
    a, b = (container_a, container_b)  # To shorten notation.
    assert type(a) == type(b)
    assert a.get_cgroup_path() == b.get_cgroup_path()
    if type(a) == ContainerSet:
        assert len(a._subcontainers) == len(b._subcontainers)
        assert all([child_a.get_cgroup_path() == child_b.get_cgroup_path()
                    for child_a, child_b in zip(list(a._subcontainers.values()),
                                                list(b._subcontainers.values()))])


def assert_equal_containers_list(containers_a: List[ContainerInterface],
                                 containers_b: List[ContainerInterface]):
    """Compares two lists (order does have importance) of containers.
       IMPORTANT: two containers are HERE
       treated as equal if they have the same cgroup_path set.

       One special assumption is made about private attribute of
       ContainerSet: that field _subcontainers exists with type
       Dict[str, Container]."""
    assert len(containers_a) == len(containers_b)
    for a, b in zip(containers_a, containers_b):
        assert_equal_containers(a, b)`

tests/test_containers.py
felidadae: Its duplicated partly. ALso checks if resgroup measurement is added. Also the real code of get_measurements is called if undefined variables or funtions are called we will catch it here.

tests/test_metrics.py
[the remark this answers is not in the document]
felidadae: fixed

tests/test_kubernetes.py
felidadae: fixed

tests/test_containers.py
felidadae: I suggest another names, which clean the whole thing:
```python
        Note: we have three variables names with the same postfix:
        * pre_running_containers - state of ContainerManager before (pre) call sync_containers_state
        * expected_running_containers - similar as above but state expected after the call,
        * got_running_containers - similar as above but state which we got after the call.
        All of three are of the same type Dict[Task, ContainerInterface].
```

tests/test_containers.py
felidadae: the same as above

owca/testing.py
```diff
@@ -121,15 +122,33 @@ def task(cgroup_path, labels=None, resources=None, subcgroups_paths=None):
...
-def container(cgroup_path, resgroup_name=None, with_config=False):
-    """Helper method to create container with patched subsystems."""
-    with patch('owca.containers.ResGroup'), patch('owca.containers.PerfCounters'):
-        return Container(
-            cgroup_path,
-            rdt_enabled=False, platform_cpus=1,
-            allocation_configuration=AllocationConfiguration() if with_config else None,
-            resgroup=ResGroup(name=resgroup_name) if resgroup_name is not None else None
-        )
+def container(cgroup_path, subcgroups_paths=[], resgroup_name=None,
```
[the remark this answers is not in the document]
felidadae: i would need to do the same with task, metric and others to keep convention. I suggest to keep that in mind, discuss that and do it in another PR.

owca/resctrl.py
felidadae: its not part of this PR now

owca/kubernetes.py
felidadae: fixed

tests/test_containers.py
```diff
@@ -12,101 +12,221 @@
...
+from owca.cgroups import Cgroup
+from owca.perf import PerfCounters
+from owca.resctrl import ResGroup
 from owca.testing import task, container
 from owca.allocators import AllocationConfiguration
 
 
+def assert_equal_containers(container_a: ContainerInterface,
+                            container_b: ContainerInterface):
+    a, b = (container_a, container_b)  # To shorten notation.
```
iwankgb: Rename arguments, please. It is perfectly clear what they represent thanks to type hinting.

tests/test_containers.py
iwankgb: Documentation is not fixed.

tests/test_containers.py
```diff
@@ -12,101 +12,221 @@
...
+    measurements = containerset.get_measurements()
+
+    resgroup_mock.get_measurements.assert_called_once()
+    assert {'foo': 3, 'cgroup_metric__1': _ANY_METRIC_VALUE * N,
+            'perf_event_metric__1': _ANY_METRIC_VALUE * N} == measurements
+
+
+def _smart_get_pids():
+    # Note: here List[int] is used instead of >>int<<
+    #   to pass mutable object (Integers are inmutable in Python).
```
iwankgb:
```suggestion
    #   to pass mutable object (Integers are immutable in Python).
```

tests/test_containers.py
iwankgb:
```suggestion
def test_container_get_allocations(*mock):
```
(it is unclear whether you test `Container` or `ContainerSet` to me).

owca/containers.py
```diff
@@ -53,81 +53,248 @@ def _sanitize_cgroup_path(cgroup_path):
...
+
+    @abstractmethod
+    def get_allocations(self) -> TaskAllocations:
+        ...
+
+
+class ContainerSet(ContainerInterface):
+    def __init__(self,
+                 cgroup_path: str, cgroup_paths: List[str], platform_cpus: int,
+                 allocation_configuration: Optional[AllocationConfiguration] = None,
```
iwankgb:
```suggestion
                 allocation_configuration: AllocationConfiguration = None,
```
(https://docs.python.org/3.6/library/typing.html#typing.Optional)

owca/containers.py
iwankgb: @ppalucki, @felidadae I would like to have a quick chat about this issue.

tests/test_containers.py
```diff
@@ -12,101 +12,221 @@
...
+        ([], {'/t1': '/t1', '/t2': '/t2'}, {},
+         {}),
     )
+)
+def test_sync_containers_state(get_pids_mock, sync_mock, perf_counters_mock,
+                               add_pids_mock, clean_taskless_groups_mock,
+                               subcgroups,
+                               tasks_, pre_running_containers_,
+                               mon_groups_relation, expected_running_containers_):
+    """Test both Container and ContainerSet classes.
```
iwankgb:
```suggestion
    """Tests both Container and ContainerSet classes.
```

tests/test_containers.py
```diff
@@ -12,101 +12,221 @@
...
-    with patch('owca.resctrl.read_mon_groups_relation', return_value=mon_groups_relation):
-        got_containers = containers_manager.sync_containers_state(tasks)
+        Note: the input arguments >>tasks_, existing_containers_, expected_running_containers_
+        contains in their names underscore on the end to distinguish them from the ones
+        created inside the function body but stress the correspondence: the input arguments
+        are used to create real objects. We cannot pass already created objects, as to
+        create them we need another argument from first of two paramatrize decorators
+        >>subcgroups<<.
+
+        Note: we have three variables names with the same postfix:
```
iwankgb: You should chose remove this paragraph or the one above, I think.

tests/test_containers.py
```diff
@@ -12,101 +12,221 @@
...
+    """Check whether summing of metrics for children containers are done properly.
+       Note: because we are mocking here classes from which measurements are read,
+       to calculate the proper value of ContainerSet we just need to multiple that
+       single value by count of subcontainers (here defined as N)."""
+    N = 3  # 3 subcontainers are created.
+    subcgroups_paths = ['/t1/c1', '/t1/c2', '/t1/c3']
+    containerset = container('/t1', subcgroups_paths, should_patch=False, rdt_enabled=True)
+
+    resgroup_mock = Mock(spec=ResGroup, get_measurements=Mock(return_value={'foo': 3}))
+    containerset.set_resgroup(resgroup=resgroup_mock)
```
iwankgb: Could you, for the sake of consistency and readability, move the mock to annotations, please?

owca/containers.py
felidadae: Default value is None. It should stay as Optional: AllocationConfiguration or None.

owca/testing.py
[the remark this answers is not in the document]
felidadae: https://jira.devtools.intel.com/browse/WC1-225

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
felidadae: fixed

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+                               add_pids_mock, clean_taskless_groups_mock,
+                               subcgroups,
+                               tasks_, pre_running_containers_,
+                               mon_groups_relation, expected_running_containers_):
+    """Tests both Container and ContainerSet classes.
 
-    # Prepare internal state used by sync_containers_state function - mock.
-    # Use list for copying to have original list.
-    containers_manager.containers = dict(existing_containers)
+        Note: the input arguments >>tasks_, existing_containers_, expected_running_containers_
```
iwankgb:
```suggestion
        Note: the input arguments tasks_, existing_containers_, expected_running_containers_
```

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+        contains in their names underscore on the end to distinguish them from the ones
```
iwankgb:
```suggestion
        contain in their names underscore at the end to distinguish them from the ones
```

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+        created inside the function body but stress the correspondence: the input arguments
```
iwankgb:
```suggestion
        created inside the function body to emphasize the relationship: the input arguments
```

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+        are used to create real objects. We cannot pass already created objects, as to
```
iwankgb:
```suggestion
        is used to create real objects. We cannot pass already created objects, as to
```

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+        create them we need another argument from first of two paramatrize decorators
```
iwankgb:
```suggestion
        create them we need another argument from first of two paramatrize decorators:
```

tests/test_containers.py
```diff
@@ -12,101 +12,217 @@
...
+        >>subcgroups<<.
```
iwankgb:
```suggestion
        subcgroups.
```

tests/test_containers.py
iwankgb: Problem is still explained twice. Leave the list below or the paragraph above, but not both.

felidadae: They explain two different things. 1)Why underscore 2) common types of variables.

PR #63, opened 2019-03-05
Allows to add additional labels to all newly created workloads

iwankgb:


workloads/run_workloads_inventory.template.yaml
```diff
@@ -43,6 +43,9 @@ application_hosts:
     wrapper_kafka_brokers: 10.10.10.99
     docker_registry: 10.10.10.99:80
 
+    # Allows to add additional labels to all newly created workloads.
+    additional_labels: '{"key":"value"}'
```
felidadae: why use json, rather than yaml dictionaries?
additional_labels:
   key: value
   key2: value2

felidadae: why use json, rather than yaml dictionaries?
additional_labels:
   key: value
   key2: value2

workloads/run_workloads.yaml
```diff
@@ -61,7 +61,7 @@
     env_uniq_id: "{{env_uniq_id}}"
     image_name: "{{image_name}}"
     image_tag: "{{image_tag_}}"
-    labels: "{{workload_version.labels | default({}) | to_json }}"
+    labels: "{{workload_version.labels | default({}) | combine(additional_labels | default('{}') | from_json) | to_json }}"
```
felidadae: why not:
``{{workload_version.labels |  combine(additional_labels) | default({}) | to_json }}``
together with using yaml dictionaries rather than json?

workloads/run_workloads_inventory.template.yaml
iwankgb: Because it allows to specify the labels in CLI, eg: `-e additional_labels='{"key":"value"}'` and it won't be possible using yaml.

PR #66, opened 2019-03-11
Update dependencies to latest available versions as of 2019.3.11

ppalucki:
All versions use exact match and are defined in:
- **Pipfile** - for building develop environment and installing tox
- **tox.ini** - to run quality checkers, unit tests and create pex based distributions
- **setup.py** - to define requirements, that are used by pex to create pex file


tox.ini
```diff
@@ -6,18 +6,23 @@ basepython = python3.6
...
-deps = flake8==3.6.0
+deps = 
+    flake8==3.7.7
 commands = flake8 owca tests example workloads
 
 [testenv:unit]
 usedevelop = True
-deps = pytest-cov
+deps = 
+    pytest==3.9.1
```
felidadae: above is 4.3.0

ppalucki: fixed

PR #67, opened 2019-03-12
Keep all dependencies specification in single place.

ppalucki:
Cons:
- some magic logic in setup.py to workaround how pex uses setup.py (it is read and send on std input for python, and those files are missing, but we provide requirements explicitly.
- for some tox virtualenvs there is more libs installed that nessesary, e.g. flake tests installs pex as well to be consistent with owca_package or wrapper_package targets

To keep *requriments.txt files with Pipfile and Pipfile.lock one needs to:
```bash
pipenv install -r requirements.txt && pipenv install --dev -r build_requirements.txt
```


ppalucki:
superseded by #70

PR #72, opened 2019-03-13
Adding contending_workload_instance label to anomaly metrics

iwankgb:


tests/test_runner.py
```diff
@@ -157,7 +173,7 @@ def test_runner_containers_state(*mocks):
         detect=Mock(
             return_value=(
                 [anomaly(
-                    'task1', ['task2'], metrics=[
+                    'task-id-/t1', ['task-id-/t2'], metrics=[
```
iwankgb: These two IDs were inconsistent with `task()` helper funciton.

owca/testing.py
```diff
@@ -65,22 +66,25 @@ def __getitem__(self, path):
...
-                type=MetricType.COUNTER
-            ))
+        metric = Metric(name='anomaly', value=1,
+                        labels=dict(contended_task_id=contended_task_id, contending_task_id=task_id,
+                                    resource=ContendedResource.MEMORY_BW, uuid=uuid,
+                                    type='contention',
+                                    contending_workload_instance=contending_workload_instances[
+                                        task_id], workload_instance=contending_workload_instances[
+                                contended_task_id]), type=MetricType.COUNTER)
+        if contended_task_id in labels:
```
iwankgb: We have nested control structures in `anomaly_metrics()` and I had to add two arguments to the function.

tests/test_runner.py
ppalucki: FYI: it was fixed in allocation branch here, and now looks like [this](https://github.com/ppalucki/owca/blob/allocation-wip-v2/owca/testing.py#L108)
and tests are fixed as [well](https://github.com/ppalucki/owca/blob/allocation-wip-v2/tests/runners/test_runners_detection.py#L130)


```python
def task(cgroup_path, labels=None, resources=None):
    prefix = cgroup_path.replace('/', '')
    return Task(
        cgroup_path=cgroup_path,
        name=prefix + '_tasks_name',
        task_id=prefix + '_task_id',
        labels=labels or dict(),
        resources=resources or dict()
    )
```

PR #73, opened 2019-03-14
Check metrics exists in list of metrics include labels submatch tooling.

ppalucki:
two functions:
- `assert_metric` - metrics with matching labels existing in given list and value optionally equals (used for testing for some metrics in `storage.store` calls)
- `assert_subdict` - on dict is a subdict of another in recursive manner (used for testing that `detect`/`allocate` got right data

### DIFF
--- a/owca/testing.py
+++ b/owca/testing.py
@@ -16,7 +16,7 @@
 """Module for independent simple helper functions."""
 
 import os
-from typing import List, Dict, Union
+from typing import List, Dict, Union, Optional
 from unittest.mock import mock_open, Mock
 
 from owca.detectors import ContendedResource, ContentionAnomaly, _create_uuid_from_tasks_ids
@@ -109,3 +109,58 @@ def task(cgroup_path, labels=None, resources=None):
         labels=labels or dict(),
         resources=resources or dict()
     )
+
+ 
+def assert_subdict(got_dict: dict, expected_subdict: dict):
+    """Assert that one dict is a subset of another dict in recursive manner.
+    """
+    for expected_key, expected_value in expected_subdict.items():
+        if expected_key not in got_dict:
+            raise AssertionError('key %r no found in %r' % (expected_key, got_dict))
+        got_value = got_dict[expected_key]
+        if isinstance(expected_value, dict):
+            # for dict use subsets
+            if isinstance(got_value, dict):
+                raise AssertionError('expected dict type at %r key, got %r' % (
+                    expected_key, type(got_value)))
+            assert_subdict(got_value, expected_value)
+        else:
+            # for other types check ordinary equality operator
+            assert got_value == expected_value
+
+
+def _is_dict_match(got_dict: dict, expected_subdict: dict):
+    """Match values and keys from dict (non recursive)."""
+    for expected_key, expected_value in expected_subdict.items():
+        if expected_key not in got_dict:
+            return False
+        else:
+            if got_dict[expected_key] != expected_value:
+                return False
+    return True
+
+
+def assert_metric(got_metrics: List[Metric],
+                  expected_metric_name: str,
+                  expected_metric_some_labels: Optional[Dict] = None,
+                  expected_metric_value: Optional[Union[float, int]] = None,
+                  ):
+    """Assert that given metrics exists in given set of metrics."""
+    found_metric = None
+    for got_metric in got_metrics:
+        if got_metric.name == expected_metric_name:
+            # found by name, should we check labels ?
+            if expected_metric_some_labels is not None:
+                # yes check by labels
+                if _is_dict_match(got_metric.labels, expected_metric_some_labels):
+                    found_metric = got_metric
+                    break
+            else:
+                found_metric = got_metric
+                break
+    if not found_metric:
+        raise AssertionError('metric %r not found' % expected_metric_name)
+    # Check values as well
+    if expected_metric_value is not None:
+        assert found_metric.value == expected_metric_value, 'metric value differs'
+
--- a/owca/testing.py
+++ b/owca/testing.py
@@ -113,19 +113,21 @@ def task(cgroup_path, labels=None, resources=None):
 
 def assert_subdict(got_dict: dict, expected_subdict: dict):
     """Assert that one dict is a subset of another dict in recursive manner.
+    Check if expected key exists and if value matches expected value.
     """
     for expected_key, expected_value in expected_subdict.items():
         if expected_key not in got_dict:
             raise AssertionError('key %r not found in %r' % (expected_key, got_dict))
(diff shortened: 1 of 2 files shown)


felidadae:
Like the idea.

tests/test_testing.py
```diff
@@ -21,3 +37,70 @@ def test_create_open_mock_manually():
...
+    (dict(x=1), dict(), True),  # ditto,
+    (dict(), dict(x=1), False),  # Non empty query doesn't match empty dict,
+    (dict(x=2), dict(x=1), False),  # no when value is different
+    (dict(x=1), dict(x=1), True),  # ok for the same
+    (dict(x=1, y=2), dict(x=1), True),  # additional parameters doesn't interfere
+    (dict(x=1, y=2), dict(x=1, y=2), True),  # two params exact match
+    (dict(x=1, y=1, z=3), dict(x=1, y=2), False),  # two keys mismatch
+    (dict(x=1, y=2, z=3), dict(x=1, y=2), True),  # two vs three keys match
+])
+def test_match_keys(got_dict, expected_subdict, expected_match):
```
iwankgb: This is not about keys; rename the function, please.

owca/testing.py
```diff
@@ -109,3 +109,59 @@ def task(cgroup_path, labels=None, resources=None):
...
+
+def assert_subdict(got_dict: dict, expected_subdict: dict):
+    """Assert that one dict is a subset of another dict in recursive manner.
+    """
+    for expected_key, expected_value in expected_subdict.items():
+        if expected_key not in got_dict:
+            raise AssertionError('key %r not found in %r' % (expected_key, got_dict))
+        got_value = got_dict[expected_key]
+        if isinstance(expected_value, dict):
+            # for dict use subsets
```
iwankgb: I can't understand your comment. Could you rephrase it?

owca/testing.py
```diff
@@ -109,3 +109,59 @@ def task(cgroup_path, labels=None, resources=None):
...
+            if not isinstance(got_value, dict):
+                raise AssertionError('expected dict type at %r key, got %r' % (
+                    expected_key, type(got_value)))
+            assert_subdict(got_value, expected_value)
+        else:
+            # for other types check ordinary equality operator
```
iwankgb: We don't care about lists, do we?

owca/testing.py
```diff
@@ -109,3 +109,59 @@ def task(cgroup_path, labels=None, resources=None):
...
+                'value differs got=%r expected=%r at key=%r' % (
+                    got_value, expected_value, expected_key)
+
+
+def _is_dict_match(got_dict: dict, expected_subdict: dict):
+    """Match values and keys from dict (non recursive)."""
+    for expected_key, expected_value in expected_subdict.items():
+        if expected_key not in got_dict:
+            return False
+        else:
```
iwankgb:
```suggestion
        else if got_dict[expected_key] != expected_value:
```

owca/testing.py
```diff
@@ -109,3 +109,59 @@ def task(cgroup_path, labels=None, resources=None):
...
+            if got_dict[expected_key] != expected_value:
```
iwankgb:
```suggestion
```

owca/testing.py
```diff
@@ -109,3 +109,59 @@ def task(cgroup_path, labels=None, resources=None):
...
+                return False
```
iwankgb:
```suggestion
            return False
```

tests/test_testing.py
```diff
@@ -21,3 +37,70 @@ def test_create_open_mock_manually():
...
+        (dict(x=1, y=dict(z=1)), dict(x=1, y=dict()), None),
+        (dict(x=1, y=dict(z=1)), dict(x=1, y=dict(z=1)), None),
+        (dict(x=1, y=dict(z=1)), dict(y=dict(z=1)), None),
+        # some deep value differs
+        (dict(x=1, y=dict(z=1)), dict(x=1, y=dict(z=2)), "value differs"),
+        (dict(x=1, y=dict(z=1)), dict(y=dict(not_existient_key=2)), "not found"),
+    ])
+def test_assert_subdict(got_dict, expected_subdict, exception_message):
+    if exception_message is not None:
+        with pytest.raises(AssertionError, match=exception_message):
```
iwankgb: Could you extend `exception_message` a bit (to make sure that the exception is raised because of particular value). String "value differs" is very generic and will match a lot of cases, not necessarily the one that we want to match.

tests/test_testing.py
```diff
@@ -21,3 +37,70 @@ def test_create_open_mock_manually():
...
+    'got_dict, expected_subdict, exception_message', [
+        # propery empty or simple dicts
+        (dict(), dict(), None),
+        (dict(x=1), dict(), None),
+        (dict(x=1), dict(), None),
+        (dict(x=1, y=1), dict(x=1), None),
+        # invalid flat dicts
+        (dict(x=1, y=1), dict(x=1, z=2), "key 'z' not found"),
+        (dict(x=1, y=1), dict(x=1, y=2), "value differs"),
+        (dict(x=1, y=1), dict(x=1, y=2), "value differs"),
```
iwankgb:
```suggestion
```
(identical to the case above).

owca/testing.py
ppalucki: no we don't - I don't not want implement and we don't need to use list to check a list is contained by another list (assert_subdict is only for dicts)

tests/test_testing.py
ppalucki: the real exception message is "verbose" and tells you everything you need, eg.
'value differs got=%r with expected=%r at key=%r' 
only the match is simple because, because it is regexp and you need to escape all the special characters and makes the test unreadable

tests/test_testing.py
ppalucki: fixed - not it is test_is_dict_match

owca/testing.py
ppalucki: Rephrased to:
```If expected value is a dict, call assert_subdict recursively.```

owca/testing.py
ppalucki: fixed - removed else entirely (because `return statement` already ends processing)

owca/testing.py
ppalucki: fixed, explanation above

owca/testing.py
ppalucki: fixed, explanation above

tests/test_testing.py
ppalucki: fixed - now comment is much better explaining what was expected and what we got (and where)

tests/test_testing.py
ppalucki: fixed

PR #84, opened 2019-03-28
Kubernetes support.

felidadae:


.env
```diff
@@ -0,0 +1 @@
+PYTHONPATH=.
```
iwankgb: What is the purpose of this file?

owca/kubernetes.py
```diff
@@ -133,22 +133,32 @@ def get_tasks(self) -> List[Task]:
                 subcgroups_paths=containers_cgroups))
 
         _log_found_tasks(tasks)
+
+        if not _are_all_tasks_of_single_qos(tasks):
+            log.warning("Tasks from namespaces {} are not all in the same qos groups."
```
iwankgb: We have discussed it today: tasks can be of various QoS class but cgroup structure must be flat.

owca/kubernetes.py
```diff
@@ -133,22 +133,32 @@ def get_tasks(self) -> List[Task]:
...
-        return os.path.join('/kubepods',
-                            '' if qos == 'guaranteed' else qos,
-                            'pod{}'.format(pod_id),
-                            container_id, "")
+        result = os.path.join('/kubepods',
+                              '' if qos == 'guaranteed' else qos,
+                              'pod{}'.format(pod_id),
+                              container_id, "")
+    # Remove last slash from path.
+    if len(result) > 1 and result[-1] == '/':
```
iwankgb: Why have you decided to change the logic? It was discussed multiple times and such change will probably affect workloads running on Mesos.

tests/test_containers.py
```diff
@@ -230,3 +230,11 @@ def test_containerset_get_pids(*args):
...
     c = container("/t1", [], rdt_enabled=True, resgroup_name='t1')
     assert c.get_allocations() == {}
+
+
+@patch('owca.containers.ResGroup.get_allocations', return_value={})
+@patch('owca.cgroups.Cgroup.get_allocations', return_value={})
+@patch('owca.containers.ContainerSet', spec=ContainerSet)
+def test_containerset_get_allocations(*mock):
+    c = container("/t1", ['/t1/s1', '/t1/s2'], rdt_enabled=True, resgroup_name='t1')
+    assert c.get_allocations() == {}
```
iwankgb: Increase number of tested scenarios, please.

owca/kubernetes.py
felidadae: thats the logic which is contained in mesos module, i know the fact that previously i added the slash on the end can be misleading.

.env
felidadae: when running pipenv shell this file is automatically sourced

owca/kubernetes.py
felidadae: As we discussed today I will only check if all tasks are in the same qos group while allocations is going be be applied which change cpu.shares.

.env
felidadae: I will remove this file from this PR, as it has no relation

owca/kubernetes.py
felidadae: https://github.com/ppalucki/owca/blob/allocation-wip-v2/owca/containers.py#L48
here we make assuption than path ends without a slash

configs/kubernetes_example_allocator.yaml
```diff
@@ -0,0 +1,36 @@
+# Copyright (c) 2019 Intel Corporation
```
ppalucki: rebase on allocation-runner and move to proper kubernetes folder

owca/runners/allocation.py
```diff
@@ -112,6 +114,20 @@ def rdt_allocation_value_constructor(rdt_allocation: RDTAllocation,
...
+    # Ignore if not KubernetesNode.
+    if not have_tasks_qos_label(tasks):
+        return
+
+    if not are_all_tasks_of_single_qos(tasks):
+        for task_id, allocation in allocations.items():
+            if AllocationType.SHARES in allocation:
+                raise InvalidAllocations('not all tasks are of the same Kubernetes QOS class '
+                                         'and at least one of the allocation contains '
+                                         'cpu share.')
```
ppalucki: Could you add/replace to comment something like this:
- "Mixing QoS classes and shares allocation is not supported! Please reschedule Pod using with the same QoS class for shares allocation."

felidadae: Fixed, but I did not add part  `Please reschedule Pod using with the same QoS class for shares allocation.` as imho it is in contradiction with the fact that OWCA is still running.

tests/test_containers.py
felidadae: I mock all called functions inside get_allocations. All the logic branches are tested here.

owca/kubernetes.py
```diff
@@ -0,0 +1,232 @@
...
+    #   Reference in source code for kubernetes version stable 1.13:
+    #   https://github.com/kubernetes/kubernetes/blob/v1.13.3/pkg/kubelet/cm/cgroup_manager_linux.go#L207
+
+    cgroup_driver: CgroupDriverType = field(
+        default_factory=lambda: CgroupDriverType(CgroupDriverType.CGROUPFS))
+
+    # By default use localhost, however kubelet may not listen on it.
+    kubelet_endpoint: str = 'https://127.0.0.1:10250'
+
+    # Key and certificate to access kubelet API.
```
ppalucki: is it required or optional ?

owca/resctrl_allocations.py
```diff
@@ -324,6 +325,28 @@ def _parse_schemata_file_row(line: str) -> Dict[str, str]:
     return domains
 
 
+def _compare_rdt_allocation_l3(current: Optional[str], new:Optional[str]):
+    # In case when at least one of the allocations is None return True.
```
ppalucki: add doc string

ppalucki: used that for mb as well

owca/resctrl_allocations.py
```diff
@@ -206,9 +206,10 @@ def calculate_changeset(self, current: Optional['RDTAllocationValue']):
             target = current._copy(target_rdt_allocation)
 
             # Prepare changeset
-            # Logic: if new value exists and is different from old one the use new.
+            # Logic: if new value exists and is different from old one use the new.
             if new.rdt_allocation.l3 is not None \
-                    and current.rdt_allocation.l3 != new.rdt_allocation.l3:
+                    and not _compare_rdt_allocation_l3(current.rdt_allocation.l3,
+                                                       new.rdt_allocation.l3):
```
ppalucki: use compare for mb string as well

owca/resctrl_allocations.py
```diff
@@ -324,6 +325,28 @@ def _parse_schemata_file_row(line: str) -> Dict[str, str]:
...
+        return first.lstrip('0') == second.lstrip('0')
+
+    # Leading underscore indicates that the data is parsed.
+    current_ = _parse_schemata_file_row(current)
+    new_ = _parse_schemata_file_row(new)
+
+    assert set(new_.keys()) & set(current_.keys()) == set(new_.keys()), \
+           "Assume that new allocations domain is subset of current allocations domain."
+
+    for domain in set(current_.keys()) & set(new_.keys()):
```
ppalucki: do it once,

owca/resctrl_allocations.py
```diff
@@ -324,6 +325,28 @@ def _parse_schemata_file_row(line: str) -> Dict[str, str]:
...
+    # In case when at least one of the allocations is None return True.
+    if current is None or new is None:
+        assert (current is None and new is None) or (new is None)
+        return True
+
+    def _compare_ignoring_leading_zeros(first, second):
+        return first.lstrip('0') == second.lstrip('0')
+
+    # Leading underscore indicates that the data is parsed.
+    current_ = _parse_schemata_file_row(current)
```
ppalucki: current_domains

owca/resctrl_allocations.py
```diff
@@ -324,6 +325,28 @@ def _parse_schemata_file_row(line: str) -> Dict[str, str]:
     return domains
 
 
+def _compare_rdt_allocation_l3(current: Optional[str], new:Optional[str]):
```
ppalucki: _is_rdt_allocation_equal

owca/resctrl_allocations.py
```diff
@@ -324,6 +325,28 @@ def _parse_schemata_file_row(line: str) -> Dict[str, str]:
...
+    # In case when at least one of the allocations is None return True.
+    if current is None or new is None:
+        assert (current is None and new is None) or (new is None)
+        return True
+
+    def _compare_ignoring_leading_zeros(first, second):
```
ppalucki: _is_equal

owca/testing.py
```diff
@@ -181,14 +201,17 @@ def run(self):
     ))
 
 
-def redis_task_with_default_labels(task_id):
+def redis_task_with_default_labels(task_id, subcgroups_paths=[]):
```
ppalucki: use immultable, and then `or []`

owca/kubernetes.py
```diff
@@ -0,0 +1,232 @@
...
+    def has_value(cls, value):
+        return any(value == item.value for item in cls)
+
+
+@dataclass
+class KubernetesNode(Node):
+    # We need to know what cgroup driver is used to properly build cgroup paths for pods.
+    #   Reference in source code for kubernetes version stable 1.13:
+    #   https://github.com/kubernetes/kubernetes/blob/v1.13.3/pkg/kubelet/cm/cgroup_manager_linux.go#L207
+
```
ppalucki: add full docstring with Args:

ppalucki: add refernce in /README.rst

owca/resctrl_allocations.py
felidadae: fixed

owca/resctrl_allocations.py
felidadae: fixed

owca/resctrl_allocations.py
felidadae: fixed

owca/resctrl_allocations.py
felidadae: fixed

owca/resctrl_allocations.py
felidadae: _does_reperform_rdt_suballocation fixed

owca/resctrl_allocations.py
felidadae: fixed

owca/testing.py
felidadae: fixed

owca/kubernetes.py
felidadae: fixed, it is optional, and its working when not set

owca/kubernetes.py
felidadae: fixed

PR #92, opened 2019-04-08
Database abstraction for etcd/zookeeper.

ppalucki:
To be used as a central storage for global configuration for whole cluster.

TODO:

- [x] design API 
- [x] implement local file based backend
- [x] implement Zookeeper based backend
- [x] implement etcd based backend 
- [x] unit tests
- [ ] perform manual integration tests with filebased/etcd/zookeeper

### DIFF
--- a/Pipfile
+++ b/Pipfile
@@ -10,6 +10,7 @@ colorlog = "==4.0.2"                  # 2018.12.14
 logging-tree = "==1.8"                # 2018.08.05
 dataclasses = "==0.6"                 # 2018.05.18
 confluent-kafka = "==1.0.0rc1"        # 2018.10.31
+kazoo = "==2.6.1"                     # 2019.01.22
 
 [dev-packages]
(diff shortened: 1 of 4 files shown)


thuang6:
it look nice, but not sure if the watcher function can be provided as well, with this functionality, each PRM agent does not need periodical pulling of new models, just pull model when it is updated from central model build service

felidadae:
When ready to review please add description how to test it (whether it can be locally or need to be run on server) and how should config file looks and how shell command should look like.

Creatone:
Example script for manual testing Etcd part:
`
from owca import databases

key = 'zm9v'
val = 'ymfy'

db = databases.EtcdDatabase(['http://localhost:2379'])

db.set(key, val)

data = db.get(key)
assert data == val
`
You can run Etcd with Docker, I used this: `https://hub.docker.com/r/xieyanze/etcd3`

PR #111, opened 2019-04-23
Upgrade urllib3

Damenus:
Due to security, it is necessary to update the package


felidadae:
Please check if its working on the production on one of the servers.

PR #118, opened 2019-04-29
Fix in etcd database

Creatone:


owca/databases.py
```diff
@@ -127,12 +128,14 @@ def set(self, key, value):
                         verify=self.ssl_verify,
                         cert=(self.client_cert_path, self.client_key_path))
                 r.raise_for_status()
-                return
+                response_data = r.json()
+                break
```
mmucek95: Why do we need `break` here?

owca/databases.py
```diff
@@ -153,10 +156,13 @@ def get(self, key):
...
 
-        if response_data:
-            if int(response_data['count']) > 0:
+        if not response_data:
+            raise Exception('EtcdDatabase: Cannot get key "{}": Timeout on all hosts!'.format(key))
+
+        if 'kvs' in response_data:
+            if 'value' in response_data['kvs'][0]:
                 return response_data['kvs'][0]['value']
             else:
```
mmucek95: Do we need this else and the one below it? I would just return None an the end of the method

PR #123, opened 2019-05-07
Replace sum of metrics for pod containers with merge and avg for derived metrics

ppalucki:


owca/metrics.py
```diff
@@ -154,29 +158,34 @@ def create_metric_with_metadata(name, value, labels=None):
...
+        if metric_name in METRICS_METADATA:
+
+            if METRICS_METADATA[metric_name].type == MetricType.GAUGE:
+                operation = lambda values: sum(values) / len(values)  # noqa
+            else:
+                assert METRICS_METADATA[metric_name].type == MetricType.COUNTER
+                operation = sum
+
+        else:
+            log.debug('defults for sum for merge operations for metric: %r', metric_name)
```
mmucek95: This log seems unclear to me.

ppalucki: The idea here was that, for any unknown metric we fallback to sum operator for merging, I'm gonna fix the log message.

Issue #129, opened 2019-05-09
not writable in local database


yiliyl:
https://github.com/intel/workload-collocation-agent/blob/7e0b939cf3a1b2f1a34158290c331f8195118297/wca/databases.py#L65-L69

I think it should be

```py
with open(full_path,'w') as f:
```

yiliyl:
@Creatone 
I think we should check if the directory exists, then create
https://github.com/intel/workload-collocation-agent/blob/7e0b939cf3a1b2f1a34158290c331f8195118297/wca/databases.py#L62-L63

btw, we shoud ensure the path before set key-value:
```
# Ensure a path, create if necessary, otherwise there will be no-node error for zookeeper
self._client.ensure_path(full_path)
```
https://github.com/intel/workload-collocation-agent/blob/7e0b939cf3a1b2f1a34158290c331f8195118297/wca/databases.py#L93-L94

Creatone:
@yiliyl Thank you for writing this issue. There are more issues because I thought that Local and Zookeper abstraction was tested. I'll fix it today.

Creatone:
#131

Issue #128, opened 2019-05-09
validate_key error in database


yiliyl:
https://github.com/intel/workload-collocation-agent/blob/7e0b939cf3a1b2f1a34158290c331f8195118297/wca/databases.py#L47-L53

I think it should be
```py
        if character not in _VALID_FILENAME_CHARACTERS:
```

Creatone:
#131

PR #131, opened 2019-05-10
Database abstraction fix

Creatone:
It fixes: #129 #128 and many other


yiliyl:
@Creatone  
for etcd database,
https://github.com/intel/workload-collocation-agent/blob/726db37eadbaf1aeb7c9c8d4d6b6e2ea893b1dcb/wca/databases.py#L151
please initialize ```api_path``` as a variable
```py
    api_path:str = '/v3alpha'
```
so for etcd 3.3.x version, we can use the api_path '/v3beta' or '/v3'

PR #140, opened 2019-05-16
MBA allocations validate fix

Creatone:
MBA allocations are not properly validated 
(/sys/fs/resctrl/info/MB/min_bandwidth) is not taken into consideration

### DIFF
--- a/wca/resctrl_allocations.py
+++ b/wca/resctrl_allocations.py
@@ -239,6 +239,7 @@ class RDTAllocationValue(AllocationValue):
             if not self.rdt_information.rdt_cache_control_enabled:
                 raise InvalidAllocations('Allocator requested RDT cache allocation but '
                                          'RDT cache control is not enabled!')
+
             validate_l3_string(self.rdt_allocation.l3,
                                self.platform_sockets,
                                self.rdt_information.cbm_mask,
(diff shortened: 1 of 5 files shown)


wca/resctrl_allocations.py
```diff
@@ -385,13 +387,26 @@ def validate_l3_string(l3, platform_sockets, rdt_cbm_mask, rdt_min_cbm_bits):
...
+
     domains = _parse_schemata_file_row(mb)
     _validate_domains(domains, platform_sockets)
 
+    for mb_value in domains.values():
+        check_mb_value(mb_value, mb_min_bandwidth)
+
+
+def check_mb_value(mb_value: str, mb_min_bandwidth):
+    mb_value = int(mb_value)
```
ppalucki: handle the exception, that mb_value is not an integer - raise InvalidAllocations that values is not integer

wca/resctrl_allocations.py
```diff
@@ -385,13 +387,30 @@ def validate_l3_string(l3, platform_sockets, rdt_cbm_mask, rdt_min_cbm_bits):
...
+
+def check_mb_value(mb_value: str, mb_min_bandwidth):
+    try:
+        mb_value = int(mb_value)
+    except ValueError:
+        raise InvalidAllocations("{} is not integer format".format(mb_value))
+
+    if mb_value < mb_min_bandwidth:
+        raise InvalidAllocations(
+                "{} mb value. Requires minimum {} bandwidth"
```
felidadae:
```suggestion
                "mb allocation smaller than minimum value {}"
```
I think we dont need to print  the actual value being set

PR #141, opened 2019-05-17
Implement cpuset allocation

Creatone:
[https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt](https://www.kernel.org/doc/Documentation/cgroup-v1/cpusets.txt)


wca/cgroups_allocations.py
```diff
@@ -59,3 +60,95 @@ def generate_metrics(self):
...
+        if len(boundaries) == 1:
+            cores.add(int(boundaries[0]))
+        elif len(boundaries) == 2:
+            start = int(boundaries[0])
+            end = int(boundaries[1])
+
+            for i in range(start, end + 1):
+                cores.add(i)
+
+    return list(cores)
```
ppalucki: should be `list(sorted(cores))`

wca/cgroups.py
```diff
@@ -39,36 +35,80 @@
 QUOTA_NORMALIZED_MAX = 1.0
 
 
+class CgroupPath(str, Enum):
```
felidadae: dont like name -> cgroup path suggest that it is full path to a cgroup. here you have cgroup subsystem root dir

wca/extra/tester.py
```diff
@@ -217,13 +229,15 @@ def check(self, metrics):
             raise CheckFailed('File {} does not exist!'.format(self.path))
 
         with open(self.path) as f:
+            correct = True
```
mmucek95: Please get rid of correct variable

PR #142, opened 2019-05-21
Check that full path is used for location of config file

Damenus:


configs/extra/tester_example.yaml
```diff
@@ -1,6 +1,6 @@
 runner: !AllocationRunner
   node: &tester !Tester
-   config: 'configs/extra/tester_config.yaml'
+   config: '${pwd}/configs/extra/tester_config.yaml'
```
Damenus: I think it is hard to validate path config from yaml. And I have problem to prepare tests for this example

PR #145, opened 2019-05-28
Add generic mechanism to handle errors and warnings and save them as metrics	

not7cd:
This PR adds basic logging measurement per record severity and module.
Two metrics are exposed for severity ERROR and WARNING.


wca/logger.py
```diff
@@ -25,6 +28,38 @@
...
+        metrics.extend([
+                Metric(
+                    name="wca_warning_count",
+                    type=MetricType.GAUGE,
+                    value=counter[logging.WARNING],
+                    labels={"module": logger_name}
+                ),
+                Metric(
+                    name="wca_error_count",
+                    type=MetricType.GAUGE,
```
Creatone:
```suggestion
                    type=MetricType.COUNTER,
```

wca/logger.py
```diff
@@ -25,6 +28,38 @@
...
+        self.counters[record.name][record.levelno] += 1
+
+
+def get_logging_metrics() -> List[Metric]:
+    metrics = []
+    for logger_name, counter in _module_record_counters.items():
+        metrics.extend([
+                Metric(
+                    name="wca_warning_count",
+                    type=MetricType.GAUGE,
```
Creatone:
```suggestion
                    type=MetricType.COUNTER,
```

wca/logger.py
```diff
@@ -25,6 +28,38 @@
 
 log = logging.getLogger(__name__)
 
+_module_record_counters = defaultdict(Counter)
+
+
+class CountingHandler(logging.Handler):
+    def __init__(self, counters):
```
Creatone:
```suggestion
    def __init__(self, counters: Dict[str, Counter]):
```

wca/logger.py
```diff
@@ -25,6 +28,38 @@
...
+        super().__init__()
+        self.counters = counters
```
Creatone: Check if passed argument is valid with:
``` from wca.config import assure_type ```
```suggestion
        assure_type(counters, Dict[str, Counter])
        self.counters = counters
```

PR #152, opened 2019-06-06
Fix bug: FileNotFoundError after raises while read measurements

Damenus:
https://github.com/intel/workload-collocation-agent/issues/150


felidadae:
Please move exception handling from container
```
    def get_measurements(self) -> Measurements:
        # Cgroup measurements
        try:
            cgroup_measurements = self._cgroup.get_measurements()
        except FileNotFoundError:
            log.warning('Could not read measurements for container %s. '
                        'Probably the mesos container has died during '
                        'the current runner iteration.',
                        self._cgroup_path)
            # Returning empty measurements.
            cgroup_measurements = {}
```
to cgroups.py module and change comment to: Could not read measurements for cgroup %s'
please remove additional comment part

PR #153, opened 2019-06-10
Cleaning of project structure

felidadae:
* move wrapper package into workloads directory
* move wca/testing.py into tests/testing.py
* move wca/extra/tester.py into tetsts/tester.py

Additionaly to make it work:
* make tests directory a package
* changes in Makefile for creating pex files and running tests
   * flag -D
   * extending PYTHONPATH

Please notice that:
* tests/testing.py file looks like added in github UI, but it is just renamed, without introducing any changes


tests/tester/test_tester.py
```diff
@@ -15,12 +15,13 @@
 from wca.nodes import Task
 from wca.config import register
```
Damenus: I think, you should remove not necessary lines

Issue #160, opened 2019-06-14
etcd connection error: `urljoin` not working as expected


yiliyl:
https://github.com/intel/workload-collocation-agent/blob/52730131feebb61d310e8463ee60bb1ef353d3c4/wca/databases.py#L194-L196

for example
```
api_path = 'http://x.x.x.x.2379/v3alpha'
url = '/kv/put'
```
after ```urljoin(api_url, url)```

return ```http://x.x.x.x.2379/kv/put```

not what we expected like ```http://x.x.x.x.2379/v3alpha/kv/put```

yiliyl:
@ppalucki

ppalucki:
fixed in #161 and backported to 1.0.x branch

Issue #159, opened 2019-06-14
TypeError: isinstance() arg 2 must be a type or tuple of types


yiliyl:
TypeError: isinstance() arg 2 must be a type or tuple of types

Path is a function here

https://github.com/intel/workload-collocation-agent/blob/52730131feebb61d310e8463ee60bb1ef353d3c4/wca/databases.py#L120-L123

yiliyl:
@ppalucki

ppalucki:
fixed in #161 and backported to 1.0.x branch

Issue #158, opened 2019-06-14
AttributeError: 'SSL' object has no attribute 'get_certs'


yiliyl:
may be you want to use 'get_client_certs'

https://github.com/intel/workload-collocation-agent/blob/52730131feebb61d310e8463ee60bb1ef353d3c4/wca/databases.py#L197-L202

https://github.com/intel/workload-collocation-agent/blob/52730131feebb61d310e8463ee60bb1ef353d3c4/wca/security.py#L138-L144

@ppalucki  please fix it

ppalucki:
fixed in #161 and backported to 1.0.x branch

PR #161, opened 2019-06-14
Database fix

Creatone:
This PR fix issues:
#160 
#159 
#158

### DIFF
--- a/wca/databases.py
+++ b/wca/databases.py
@@ -23,7 +23,7 @@ from typing import Optional, List
 
 from dataclasses import dataclass
 
-from wca.config import assure_type, Numeric, _PathType, ValidationError
+from wca.config import assure_type, Numeric, ValidationError, _PathType
 from wca.security import SSL
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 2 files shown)


wca/databases.py
```diff
@@ -127,7 +126,7 @@ def __post_init__(self):
                         certfile=self.ssl.client_cert_path,
                         keyfile=self.ssl.client_key_path,
                         )
-            else:
+            elif isinstance(self.ssl.server_verify, bool):
```
mmucek95: What if self.ssl.server_verify is neither Path nor bool?

wca/databases.py
```diff
@@ -116,7 +116,8 @@ def __post_init__(self):
         from kazoo.client import KazooClient
 
         if self.ssl:
-            if isinstance(self.ssl.server_verify, _PathType):
+            if isinstance(self.ssl.server_verify, _PathType) \
```
ppalucki: it make no sense, there is no chance that yaml has `_PathType` - just leave str

ppalucki:
LGTM

PR #164, opened 2019-06-17
Perf counter scaling as metrics

not7cd:


wca/perf.py
```diff
@@ -96,31 +97,52 @@ def _parse_online_cpus_string(raw_string) -> List[int]:
     return parsed_cpus_info
 
 
+# TODO(nszulc): move to class?
```
ppalucki: make it work first, the we think about refactor, unless you have very strong argument to move that to class (re-usability, too complex, local state)

wca/perf.py
```diff
@@ -96,31 +97,52 @@ def _parse_online_cpus_string(raw_string) -> List[int]:
...
             raw_value,
             time_enabled,
             time_running
         )
+        measurements[event_names[current_event]] = measurement
+
+        scaling_factors.append(scaling_factor)
         # id is unused, but we still need to read the whole struct
         struct.unpack('q', file.read(8))[0]
+    # TODO(nszulc): another special case handling
```
ppalucki: remove TODO, and give a comment why it is special

wca/perf.py
```diff
@@ -96,31 +97,52 @@ def _parse_online_cpus_string(raw_string) -> List[int]:
...
+            aggregated_measurements[metric_name] += measurements_from_single_cpu[metric_name]
+        # TODO(nszulc): move out (derived?)
+        aggregated_measurements[MetricName.SCALING_FACTOR_MAX] = max(
+            aggregated_measurements[MetricName.SCALING_FACTOR_MAX],
+            measurements_from_single_cpu[MetricName.SCALING_FACTOR_MAX]
+        )
+        # TODO(nszulc): move out (derived?)
+        aggregated_measurements[MetricName.SCALING_FACTOR_MEAN] += measurements_from_single_cpu[
+            MetricName.SCALING_FACTOR_MEAN]
+    if aggregated_measurements[MetricName.SCALING_FACTOR_MEAN] != 0:
```
ppalucki: why this ?

PR #169, opened 2019-06-19
Change parsing labels

Damenus:


tests/e2e/Jenkinsfile
```diff
@@ -0,0 +1,150 @@
+pipeline {
```
mmucek95: Please get rid of changes from e2e branch.

PR #173, opened 2019-06-19
Use NOPAllocator instead of ExampleDetector

mmucek95:


configs/mesos/mesos_example_allocator.yaml
```diff
@@ -0,0 +1,24 @@
+runner: !AllocationRunner
+  node: !MesosNode
+    mesos_agent_endpoint: 'http://127.0.0.1:5051'
+    timeout: 5
+    ssl: !SSL
```
mmucek95: I am considering removing ssl for purpose of easy example configuration.

PR #185, opened 2019-06-26
Make confluent-kafka optional.

felidadae:
confluent-kafka-python consitute dependency for WCA. Its not needed unless Kafka is used to store metrics. In this PR the package is made optional.

The reason why we removed the package from Pipfile:
* contains bundled binary libraries which bring to us responsibility for their security vulnerabilities: (wheel many linux package zlib1.2.3)

In this PR I also made some minor improvements:
* replacing tabs with spaces in Jenkinsfiles (the intent was mixed)
* minor refactor in wrapper package, where Storage class is used (KafkaStorage inherits from Storage)

Changes for developer:
* to use KafkaStorage locally or on the cluster one needs install dependencies as guided in docs/kafka_storage.rst
* to build wca.pex using Makefile special flag needs to be passed

Imporant consequences:
* using provided in documentation way of building WCA, which use Makefile (make wca_package), now wca.pex do not contain by default confluent_kafka package; as a consequence by default KafkaStorage is not enabled; if it will be used exception will be thrown, which will exit the application
* using provided in documentation way of building wrappers, which use Makefile (make wca_wrappers), pex files will still contain confluent_kafka downloaded from PyPI
* to build wca.pex containing kafka_consumer please follow guide docs/kafka_storage.rst
* to build wca with kafka_consumer special libararies are needed
* to run wca.pex on a node, librdkafka1 dynamic library is needed


docs/install.rst
```diff
@@ -32,6 +32,35 @@ Building executable binary (distribution)
...
+
+We require to clone the confluent-kafka-python as wheel package of version 1.0.1 (the newest) included in PyPI cannot be bundled in
+pex (https://github.com/confluentinc/confluent-kafka-python/issues/482#issuecomment-490711109).
+
+To build confluent-kafka-python librdkafka library needs to be installed on the machine.
+All commands which needs to be run to build WCA pex file with **KafkaStorage** component enabled:
+
+.. code:: shell
+
+   yum install librdkafka
```
ppalucki: Please, add a link to official docs how to install librdkafka from confluent wiki for centos distribution

wca/storage.py
```diff
@@ -269,6 +272,10 @@ class KafkaStorage(Storage):
     extra_config: Dict[Str, Str] = None
 
     def __post_init__(self) -> None:
+        if confluent_kafka is None:
+            log.warning("KafkaStorage is not supported as confluent_kafka "
```
ppalucki: raise error or `ValidationError` - without module is should be fatal - cannot imaging frustration of looking for missing metrics with improperly build pex

ppalucki:
are there any changes required for build Jenkins files to make it work ?

docs/install.rst
felidadae: fixed

wca/storage.py
felidadae: fixed

docs/kafka_storage.rst
```diff
@@ -0,0 +1,81 @@
...
+    gpgkey=https://packages.confluent.io/rpm/5.2/archive.key
+    enabled=1
+    EOF
+
+    sudo yum clean all && sudo yum install -y librdkafka1
+
+
+Unsafe method
+-------------
+We strongly advise to build the confluent-kafka-python manually due to security reasons. Included in PyPi wheel manylinux package
```
mmucek95: Sentence starting with "Included in Pypi..." is unclear. What is included where? Shouldn't it be something like: "Wheel manylinux package included in PyPi contains bundled..."

docs/kafka_storage.rst
```diff
@@ -0,0 +1,81 @@
...
+contains bundled binary libraries, where some of them are outdated (e.g. zlib1.2.3 is included in the package and contains security vulnerabilities).
+Hovewer, if You want to use the package from PyPi and skip all described here steps please run:
```
mmucek95: s/skip all described here steps/skip all steps described here

setup.py
```diff
@@ -13,11 +13,20 @@
...
 import json
 
+# Taking all packages from Pipfile.lock except for:
+# * confluent_kafka, unless INCLUDE_UNSAFE_CONFLUENT_KAFKA_WHEEL is set to 'yes'.
 install_requires = ['%s%s' % (name, spec['version'])
                     for name, spec in
                     json.load(open('Pipfile.lock'))['default'].items()]
+if os.environ.get('INCLUDE_UNSAFE_CONFLUENT_KAFKA_WHEEL', 'no') == 'yes':
+    print('Warning: bundling confluent_kafka package with binary libraries, '
+          'some of them are outdated and contains security vulnerabilities.')
```
mmucek95: s/contains/contain

PR #186, opened 2019-06-26
Verify that configuration file has proper ACLs during initialization

Damenus:


wca/main.py
```diff
@@ -33,6 +34,30 @@
 log = logging.getLogger('wca.main')
 
 
+def valid_config_file(config):
+
+    user_uid = os.stat(config).st_uid
+    if user_uid != os.getuid() and user_uid != 0:
+        log.error(
+            'Error: The config \'%s\' is not valid. User is not owner config or is not root.'
```
mmucek95: Please change "User is not owner config or is not root" to "User is not owner of the config or is not root"

wca/main.py
```diff
@@ -33,6 +34,30 @@
...
+            % config)
+        exit(1)
+
+    mode = stat.S_IMODE(os.stat(config).st_mode)
+    if mode != 700 and mode != 600:
+        log.error(
+            'Error: The config \'%s\' is not valid. It has not correct proper ACLs. '
```
mmucek95: Please use either correct or proper. Not both. Maybe something like this: "It doesn't have correct ACLs. "

wca/main.py
```diff
@@ -33,6 +34,30 @@
...
+            'Only user must be able to read and write.'
```
mmucek95: s/must/should souns better IMHO

wca/main.py
Damenus: done

wca/main.py
Damenus: done

wca/main.py
Damenus: done

PR #206, opened 2019-07-12
Remove race problem when task are being terminated

mmucek95:


wca/cgroups.py
```diff
@@ -89,7 +89,7 @@ def get_measurements(self) -> Measurements:
                 cpu_usage = int(cpu_usage_file.read())
         except FileNotFoundError:
             log.warning('Could not read measurements for cgroup %s. ' % self.cgroup_path)
-            return {}
+            raise
```
ppalucki: Please create new internal Exception e.g. `MeasurementsMissingException` with description that is raced when race is datected and task should be excluded from further processing.

here it will look like:
```python
raise MissingMeasurementsException('file %s to read measurements data is not available' % e)
```

wca/resctrl.py
```diff
@@ -211,7 +211,7 @@ def _get_event_file(socket_dir, event_name):
         except FileNotFoundError:
             log.warning("Could not read measurements from rdt - ignored! "
                         "rdt group was not found (race detected)")
-            return {}
+            raise
```
ppalucki: ditto - use dedicated exception

wca/runners/allocation.py
```diff
@@ -373,7 +373,13 @@ def _iterate_body(self,
 def _get_tasks_allocations(containers) -> TasksAllocations:
     tasks_allocations: TasksAllocations = {}
     for task, container in containers.items():
-        task_allocations = container.get_allocations()
+        try:
+            task_allocations = container.get_allocations()
+        except FileNotFoundError:
```
ppalucki: to not handle builtin exception here - we should only cover know cases 

just catch `MissingMeasurementsException` - then you can use its message in log.warning

```
except MissingMeasurementsException as e:
    log.warning('One or more allocations are missing for '
                        'container {} - ignoring! (because {})'.format(container, e),)

```

wca/runners/measurement.py
```diff
@@ -235,10 +235,11 @@ def _prepare_tasks_data(containers: Dict[Task, Container]) -> \
 
     for task, container in containers.items():
         # Task measurements and measurements based metrics.
-        task_measurements = container.get_measurements()
-        if not task_measurements:
-            log.warning('there is not measurements collected for container %r - ignoring!',
-                        container)
+        try:
+            task_measurements = container.get_measurements()
+        except FileNotFoundError:
```
ppalucki: ditto use dedicated exception

PR #217, opened 2019-07-23
Task synchronization

mmucek95:
Handle exception when getting tasks was unsuccessful.


wca/runners/measurement.py
```diff
@@ -151,7 +152,13 @@ def _iterate(self):
         iteration_start = time.time()
 
         # Get information about tasks.
-        tasks = self._node.get_tasks()
+        try:
```
felidadae: could you please test that code

wca/nodes.py
```diff
@@ -22,6 +22,10 @@
 TaskId = str
 
 
+class TaskSynchornizationException(Exception):
```
felidadae: synchorn ---> synchro

ppalucki:
@mmucek95 could you cherry-pick this as well to 1.0.x branch ?

PR #227, opened 2019-07-31
Perf pmu

mmucek95:


ppalucki:
lets suspend this until I found a way to remove eval from source branch

ppalucki:
it is going to be replaced with new code without eval code

PR #264, opened 2019-09-18
Update pipfile

Creatone:


Creatone:
> why skip e2e? e2e must pass

Updated dependencies doesn't invade K8s or Mesos code. We are currently migrating Mesos cluster. E2e passed for K8s so it would work also for Mesos.

PR #275, opened 2019-10-02
Perf pmu - move from extender

mmucek95:


mmucek95:
Will be merged to devel

PR #282, opened 2019-10-09
Pref pmu devel

mmucek95:


wca/runners/measurement.py
```diff
@@ -117,13 +116,10 @@ def __init__(
             extra_labels: Dict[Str, Str] = None,
             event_names: List[str] = DEFAULT_EVENTS,
             enable_derived_metrics: bool = False,
+            enable_perf_pmu: bool = True,
```
ppalucki: this should be added to Allocation and Detection as well

wca/runners/measurement.py
```diff
@@ -229,30 +223,34 @@ def _initialize(self) -> Optional[int]:
...
-    def _init_uncore_pmu(self, enable_derived_metrics):
-        try:
-            cpus, pmu_events = _discover_pmu_uncore_imc_config(UNCORE_IMC_EVENTS)
-        except PMUNotAvailable:
+    def _init_uncore_pmu(self, enable_derived_metrics, enable_perf_pmu):
+        if enable_perf_pmu:
+            try:
+                cpus, pmu_events = _discover_pmu_uncore_imc_config(
+                    UNCORE_IMC_EVENTS)
+            except PMUNotAvailable:
```
ppalucki: we should grab this as `as e` and then present in log.warning

ppalucki: for now it is only one place (len pmus is 0) but in the future we can find other cases - and explanation for use would be helpful

PR #286, opened 2019-10-15
Adding chierarchical measurements

felidadae:
Now measurements can be also chierarchical - in other words a measurement can be identified by metric name and a set of labels.
Not polished, but ready to be merged into devel branch.


tests/test_platforms.py
```diff
@@ -146,16 +148,16 @@ def test_collect_platform_information(*mocks):
...
         node_cpus={},
         rdt_information=RDTInformation(True, True, True, True, 'fffff', '2', 8, 10, 20),
-        measurements={},
+        measurements={MetricName.CPU_USAGE_PER_CPU: {0: 100, 1: 200},
+                      MetricName.MEM_USAGE: 1337,
+                      MetricName.MEM_NUMA_FREE: {0: 1},
+                      MetricName.MEM_NUMA_USED: {0: 2}},
     )
 
+    print(got_metrics)
```
ppalucki: drop print

PR #296, opened 2019-10-25
Getting rid of null pointer dereference

iwankgb:
Fixing null pointer dereference issue identified by Coverity.

### DIFF
--- a/wca/runners/allocation.py
+++ b/wca/runners/allocation.py
@@ -327,10 +327,10 @@ class AllocationRunner(MeasurementRunner):
             log.debug('New allocations: %s', new_allocations)
             new_allocations_values = TasksAllocationsValues.create(
                 self._rdt_enabled, new_allocations, self._containers_manager.containers, platform)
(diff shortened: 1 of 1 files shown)


wca/runners/allocation.py
```diff
@@ -327,10 +327,10 @@ def _iterate_body(self,
             log.debug('New allocations: %s', new_allocations)
             new_allocations_values = TasksAllocationsValues.create(
                 self._rdt_enabled, new_allocations, self._containers_manager.containers, platform)
-            new_allocations_values.validate()
 
             # Calculate changeset and target_allocations.
             if new_allocations_values is not None:
+                new_allocations_values.validate()
```
ppalucki: we should check if `create` can return None at all, and why all our tests, never catch such case

the problem is the null check above condition is broken, because we have never experienced this kind of bug - before merging we need to check `create`

iwankgb:
@ppalucki, fixed!

PR #299, opened 2019-10-30
Numa allocator with migrate pages as allocation

ppalucki:
includes fixes:
- break the iteration for MissingMeasurementsException to only provide consistent data to plugins,
- new config and clean up the algorhitm for numa-allocator (less print, mode debugs)


wca/extra/numa_allocator.py
```diff
@@ -11,6 +12,15 @@
 
 log = logging.getLogger(__name__)
 
+def migrate_pages(task, tasks_pids, to_node):
+    # if not all pages yet on place force them to move
+    from_node = str(0 if to_node else 1)
```
kad: this will not work on 4s machines or with snc.

ppalucki: you're right :) do you think it matters if I call move multiple times from each node
or I can just `migrate PID 0-3 target_node` ? the first one is easier ;)

kad: in theory, if `to_node=2`, then  `migratepages PID all 2` should work. or `migratepages PID !2 2`

ppalucki: fixed - but we just move from every node by now - you cannot control source node - we need to create a new feature for thata

PR #303, opened 2019-11-12
Avoid redundant init variables in runners objects.

Creatone:


wca/runners/measurement.py
```diff
@@ -302,8 +318,10 @@ def _iterate(self):
             self._wait()
             return
 
-        self._iterate_body(containers, platform, tasks_measurements, tasks_resources,
-                           tasks_labels, common_labels)
+        # Inject other code.
+        if self._iterate_body_callback:
```
felidadae: is not None

wca/runners/measurement.py
```diff
@@ -205,7 +211,11 @@ def _initialize(self) -> Optional[int]:
 
         if self._rdt_enabled:
             # Resctrl is enabled and available, call a placeholder to allow further initialization.
-            rdt_initialization_ok = self._initialize_rdt()
+            if self._initialize_rdt_callback:
```
felidadae: is not None

wca/runners/measurement.py
```diff
@@ -263,6 +273,12 @@ def _init_uncore_pmu(self, enable_derived_metrics, enable_perf_uncore):
             else:
                 self._uncore_get_measurements = self._uncore_pmu.get_measurements
 
+    def _set_iterate_body_callback(self, func):
```
felidadae: needs comments

wca/runners/measurement.py
```diff
@@ -263,6 +273,12 @@ def _init_uncore_pmu(self, enable_derived_metrics, enable_perf_uncore):
...
+        self._iterate_body_callback = func
+
+    def _set_initialize_rdt_callback(self, func):
```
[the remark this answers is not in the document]
felidadae: customize rdt initialization, e.g. MBA i MBM only needs different initialization code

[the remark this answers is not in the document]
felidadae: I would also put in the file this function near to initializa_rdt function

PR #306, opened 2019-11-19
Revert deleted mutilate workload

Damenus:


examples/kubernetes/README.md
```diff
@@ -0,0 +1,211 @@
+Kubernetes
```
ppalucki: is the new file by purpose or by accident ?

PR #311, opened 2019-11-20
Simplify API for plugins 

Creatone:
Instead of task_measurements/task_labels/task_resources - should have 
```task: {measurements: {}, labels: {}}``` or other better option.

### CODE REVIEW

File: wca/runners/allocation.py [hunk 1]
```diff
@@ -284,17 +284,16 @@ def _initialize_rdt(self) -> bool:
...
         """Allocator callback body."""
 
         current_allocations = _get_tasks_allocations(containers)
 
         # Allocator callback
         allocate_start = time.time()
         new_allocations, anomalies, extra_metrics = self._allocator.allocate(
-            platform, tasks_measurements, tasks_resources, tasks_labels,
-            current_allocations)
+            platform, tasks_data, current_allocations)
```
mmucek95: Allocate method takes only platform and task_data arguments.

wca/runners/measurement.py
```diff
@@ -414,28 +410,35 @@ def _prepare_tasks_data(containers: Dict[Task, Container]) -> \
...
+        tasks_data[task.task_id] = {
+                TaskDataType.LABELS: {},
+                TaskDataType.MEASUREMENTS: {},
+                TaskDataType.RESOURCES: {}
+                }
 
-    return tasks_measurements, tasks_resources, tasks_labels
+        tasks_data[task.task_id][TaskDataType.LABELS] = task_labels
+        tasks_data[task.task_id][TaskDataType.MEASUREMENTS] = task_measurements
+        tasks_data[task.task_id][TaskDataType.RESOURCES] = task_resources
```
mmucek95: I would suggest simplifying the code to somethig like this:
tasks_data[task.task_id] = {
TaskDataType.LABELS: task_labels,
TaskDataType.MEASUREMENTS: task_measurements,
TaskDataType.RESOURCES: task_resources
}

wca/runners/measurement.py
```diff
@@ -315,16 +315,15 @@ def _iterate(self):
 
         try:
             # Tasks data
```
felidadae: we do not need that comment

wca/runners/measurement.py
```diff
@@ -413,30 +409,35 @@ def _prepare_tasks_data(containers: Dict[Task, Container]) -> \
         if TaskResource.MEM in task.resources:
             task_measurements[MetricName.MEM.value] = task.resources[TaskResource.MEM.value]
 
-        task_labels = task.labels.copy()
+        tasks_data[task.task_id] = TaskData(
```
felidadae: not better to create constructor which takes as input Task? just to discuss

wca/detectors.py
```diff
@@ -34,10 +34,19 @@
...
-TasksResources = Dict[TaskId, Dict[str, float]]
-TaskLabels = Dict[str, str]
-TasksLabels = Dict[TaskId, TaskLabels]
+
+
+TaskMeasurements = Measurements
+TaskAllocations = Dict[str, str]
+
+
+@dataclass
```
felidadae: should find better place than wca/detectors.py as that structure is bothed used in detector and allocator.

wca/runners/measurement.py
felidadae: would be more implicit that TaskData is derived from class Task

wca/detectors.py
felidadae: however i dont know if there is better place to put that

wca/runners/allocation.py
```diff
@@ -374,19 +373,19 @@ def _iterate_body(self,
         allocations_package.send(common_labels)
 
 
-def _get_tasks_allocations(containers) -> TasksAllocations:
-    tasks_allocations: TasksAllocations = {}
+def _get_tasks_allocations(containers, tasks_data: TasksData) -> TasksAllocations:
```
felidadae: Just add comment that input parameter tasks_data is modified inside function

wca/runners/measurement.py
Creatone: Discussed offline.

tests/testing.py
```diff
@@ -318,7 +334,9 @@ def _decorated_function(*args, **kwargs):
                    return_value=(platform_mock, [metric('platform-cpu-usage')], {})), \
              patch('wca.platforms.collect_topology_information', return_value=(1, 1, 1)), \
              patch('wca.security.are_privileges_sufficient', return_value=True), \
-             patch('resource.getrusage', return_value=Mock(ru_maxrss=WCA_MEMORY_USAGE)):
+             patch('resource.getrusage', return_value=Mock(ru_maxrss=WCA_MEMORY_USAGE)), \
```
felidadae: Add comment in commit that you are solving bug by the way

tests/testing.py
```diff
@@ -318,7 +334,9 @@ def _decorated_function(*args, **kwargs):
...
+             patch('wca.perf_uncore.UncorePerfCounters._open'), \
+             patch('wca.perf.PerfCounters._open'):
```
felidadae: Add comment in commit that you are solving bug by the way.

PR #297, opened 2019-10-29
Hierarchical metrics for perf(no metric aggregation)

mmucek95:

PR #321, opened 2019-11-26
Marta's perf uncore hierarchical support.

ppalucki:
Supersedes #297 
Includes:
- fixes for Pawel access check
- fix for logging configuration
- more unittests than Marta original PR #297 and some quick fixes


wca/platforms.py
```diff
@@ -645,7 +645,7 @@ def collect_platform_information(rdt_enabled: bool = True,
 def decode_listformat(value: str) -> Set[int]:
     """Parse "List Format" as describe by man cpuset(7)
 
-    can raise ValueError in case in inproper input.
+    can raise ValueError in case in improper input.
```
Creatone: Should be ``in case of improper input.``

PR #324, opened 2019-11-27
UPI metrics cont. of Marta's work

ppalucki:
supersedes #304 
Includes derived UPI bandwidht metric.


wca/metrics.py
```diff
@@ -424,7 +430,18 @@ def create_metric_with_metadata(name, value, labels=None, granularity=None):
         return metric
 
 
-Measurements = Dict[MetricName, MetricValue]
+LevelMeasurements = Dict[Union[str, int], MetricValue]
```
felidadae: no recursion, remove types or improve

Issue #320, opened 2019-11-26
Improper loggers configuration precedence. 


ppalucki:
**Describe the bug**
Command line loggers configuration is overwritten by values from config file.

**To Reproduce**
Run wca with loggers from command line `--log trace` and configure loggers in config file like this
```
loggers:
  wca: info
```
Info level is set for all loggers - contrary to to what docs says https://github.com/intel/workload-collocation-agent/blob/master/docs/logging.rst#wca-logging-configuration

**Expected behavior**
Trace logging will be enabled as described in docs.

**Context:**
- All versions are affected.

fix here: https://github.com/intel/workload-collocation-agent/pull/321/commits/985ed5fc546b750ec53b6381da996e3069f4a543

(need to be backported to master/1.0.x)

ppalucki:
we need another fix from this https://github.com/intel/workload-collocation-agent/pull/328/commits/588316e4497f20e5a8346493dd68b30a0b89bbe3

ppalucki:
backported to 1.0.x but not yet released as new version!

PR #349, opened 2019-12-09
Fix improper loggers configuration precedence for 1.0.x

Creatone:
#320 and 588316e


ppalucki:
Remember to update issues that is fixed, but not yet released, @Creatone  #320

PR #355, opened 2019-12-11
Update metrics docs and new CPU number label

ppalucki:
Better docs for metrics (some fixes and description) about columns and about common labels and added new "cpu_model_number" as default metrics and extract this code to common function called
`common_optional_labels()`


docs/metrics.rst
```diff
@@ -5,18 +5,53 @@ Available metrics
...
+
 For searchable list of metrics `metrics as csv file <metrics.csv>`_ .
 
-The "Enabled" describes if metric is enabled by default and in brackets there is information which 
-option in MeasurementRunner is responsible for configuring it.
+Common labels
+=============
 
-.. contents:: Table of Contents
+Because all WCA collected metrics are collected in context of single physical machine (host), all metrics
```
mmucek95: "collected" is repeated. I would delete first occurrence.

docs/metrics.rst
```diff
@@ -5,18 +5,53 @@ Available metrics
...
+- **Name**: is the name of metric that will be exported to Prometheus by using Prometheus exposition format but also
+  the name of the key in ``Measurements`` dict-like type used in ``Detector`` and ``Allocator`` plugins,
+- **Help**: information what metric represents and some details how metric was collected and known problems or limitations,
+- **Unit**: unit of the metric (usually seconds or bytes),
+- **Type**: only possible types are `gauge` and `counter` as described in `Prometheus metric types <https://prometheus.io/docs/concepts/metric_types/>`_.
+- **Source**: short description about mechanics that was used to collect metric,
+  for more detailed information check out `Metric sources documenation <metric_sources.rst>`_.
+- **Enabled** - column describes if metric is enabled by default and how to enable (option in ``MeasurementRunner`` responsible for configuring it. Please refer to `metrics sources documentation <metrics_sources.rst>`_
+  for more details.)
+- **Levels/Labels** - some metrics how additional dimensions (more granularity than just ``Task`` or ``Platform``) e.g. ``task_mem_numa_pages`` can be collected per NUMA node - in this case
```
mmucek95: s/how/have

ppalucki:
@felidadae is it really required - I do not want to change API if not needed and those are not only "platform metrics" - e.g. wca_version is something else

felidadae:
@ppalucki yes i know, there is wca_version but it is only one from 7, so still it is more proper name (not ideal) than current very generic saying nothing name. but argument with changing API, seems valid to me. not blocking.

PR #360, opened 2019-12-13
Tag images create by WCA with branch name and check that there is enough available PMU counters.

ppalucki:
1. During E2E tests, echa build image will have name of branch as tag
2. Upon starting WCA, perform a check to make sure that there is enough PMU counters available  (assuming 8 for non-ht and 4 for ht enabled platforms, verified on BDX, SKX and CLX)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -238,7 +238,7 @@ pipeline {
                         withCredentials([file(credentialsId: 'specjbb', variable: 'SPECJBB_TAR')]) {
                             sh '''
                             IMAGE_NAME=${DOCKER_REPOSITORY_URL}/wca/specjbb:${GIT_COMMIT}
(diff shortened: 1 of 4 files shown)


Jenkinsfile
```diff
@@ -171,10 +189,13 @@ pipeline {
                     steps {
                     sh '''
                     IMAGE_NAME=${DOCKER_REPOSITORY_URL}/wca/cassandra_stress:${GIT_COMMIT}
+                    BRANCH_IMAGE_NAME=${DOCKER_REPOSITORY_URL}/wca/cassandra_stress':${GIT_BRANCH}
```
mmucek95: Remove **'** and should pass e2e

ppalucki: thx !!!

ppalucki: saved me time ! ! !

PR #366, opened 2019-12-17
Derived metric for perf cgroup now work with dynamic hierarchy workaround

ppalucki:
Before, there was an error when trying to handle derived metrics with perf_aggregate_cpus = true - derived metrics didn't work at all.
(better docs for perf event based and scaling metrics)

Also includes fix that now all the "scaling" metrics - have proper value **1.0** and any value above indicate that multiplexing/scaling is happening (real values are estimated) - probably caused by other tool collecting using the same counters.


wca/runners/measurement.py
```diff
@@ -217,10 +217,12 @@ def __init__(
         # We had the modify levels for all metrics
         # The set proper levels based on perf_aggregate_cpus value
         if not perf_aggregate_cpus:
-            log.debug('Enabling "cpu" level for PERF_SUBSYSTEM_WITH_CGROUPS metrics.')
+            log.debug('Enabling "cpu" level for PERF_SUBSYSTEM_WITH_CGROUPS and derived metrics.')
             for metric_metadata in METRICS_METADATA.values():
                 if metric_metadata.source == MetricSource.PERF_SUBSYSTEM_WITH_CGROUPS:
                     metric_metadata.levels = ['cpu']
+                if metric_metadata.source == MetricSource.DERIVED_PERF_WITH_CGROUPS:
+                    metric_metadata.levels = ['cpu']
```
Damenus: Above is similar condition and can be merged into one.

PR #378, opened 2020-01-13
WCA use master image for monitoring deployment and enabled RDT (auto)

ppalucki:


Jenkinsfile
```diff
@@ -463,10 +464,20 @@ def test_wca_metrics_kustomize() {
...
+    check_image = sh(script: 'curl ${DOCKER_REPOSITORY_URL}/v2/wca/rpc_perf/manifests/${GIT_COMMIT} | jq .name', returnStdout: true).trim()
+    if (check_image == 'null') {
+        print('Docker images are not available!')
+        sh "exit 1"
+    }
+}
+
+def wca_image_check() {
+    print('Check if wca docker images build for this PR ${GIT_COMMIT}')
+    /* Checking only for rpc_perf */
```
mmucek95: Checking for WCA image, not rpc_perf in this case.
We could do only one method that would take a parameter for image name to avoid doubling the code.

wca/runners/measurement.py
```diff
@@ -307,12 +319,25 @@ def _initialize(self) -> Optional[int]:
...
+        # so we need access to write in cgroups.
+        write_to_cgroup = self._allocation_configuration is not None
+        use_perf = len(self._event_names) > 0
+        # Check we have enough access.
+        if not security.are_privileges_sufficient(write_to_cgroup, self._rdt_enabled, use_perf):
             return 1
 
+        # Resctrl is enabled and available, call a placeholder to allow further initialization.
+        # For MeasurementRunner it's nothing to configure in RDT to measure resource usage.
+        # Check if it's needed to specific rdt initialization in case
```
mmucek95: Did you mean: 'specify'?

ppalucki: it should be "specific" as an adjective to "functionality" keyword

mmucek95: Then the sentence should look something like this: Check if specific rdt initialization is needed in case of using...

wca/runners/measurement.py
```diff
@@ -257,46 +257,58 @@ def _wait(self):
...
+            resctrl_available,
             gather_hw_mm_topology=self._gather_hw_mm_topology
         )
-        rdt_information = platform.rdt_information
 
+        # Confront RDT (resctrl fs) with HW enabled monitoring features.
+        if self._rdt_enabled and not platform.rdt_information.is_monitoring_enabled():
+            # Note: WCA does not support RDT without monitoring (keeps a mapping of
+            # cgroups and resctrl groups).
+            msg = ('Resctrl ia available byt RDT monitoring features are not!' +
```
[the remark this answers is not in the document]
ppalucki: fixed

wca/runners/measurement.py
```diff
@@ -307,12 +319,25 @@ def _initialize(self) -> Optional[int]:
...
+        # _allocation_configuration is set in allocation mode (AllocationRunner)
+        # so we need access to write in cgroups.
+        write_to_cgroup = self._allocation_configuration is not None
+        use_perf = len(self._event_names) > 0
+        # Check we have enough access.
+        if not security.are_privileges_sufficient(write_to_cgroup, self._rdt_enabled, use_perf):
             return 1
 
+        # Resctrl is enabled and available, call a placeholder to allow further initialization.
+        # For MeasurementRunner it's nothing to configure in RDT to measure resource usage.
```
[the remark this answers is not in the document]
ppalucki: It's P. Szulik to blame here :P - I have only copied it, but I will fix that.

PR #380, opened 2020-01-14
Fix missing 'hostIP' in PodList when Pod is in Pending state.

Creatone:
Got ```KeyError: 'hostIP'``` when Pod is in "PENDING" state.


wca/kubernetes.py
```diff
@@ -213,24 +213,38 @@ def get_tasks(self) -> List[Task]:
...
+            # Lacking needed information.
+            if not container_statuses:
+                log.debug('Ignore pod with uid={} name={}. '
+                          'Pod has no information about container statuses.'
+                          .format(pod_id, pod_name))
                 continue
 
+            # Kubeapi returns all pods in cluster.
+            if not self.kubelet_enabled:
+                if host_ip is None:
```
mmucek95: You're doubling code. Maybe make one condition: if host_ip is None or host_ip != self.node_ip.strip()

Creatone: I forgot to change log message in second condition. Checkout now!

PR #393, opened 2020-02-03
Uncore metrics update

mmucek95:
* adding new uncore metrics
* removing enable_perf_uncore flag and replacing it with a list of requested events


wca/metrics.py
```diff
@@ -880,7 +884,48 @@ class MetricMetadata:
...
+        MetricMetadata(
+            'Pending queue allocations',
+            MetricType.GAUGE,
+            MetricUnit.NUMERIC,
+            MetricSource.PERF_SUBSYSTEM_UNCORE,
+            MetricGranularity.PLATFORM,
+            ['socket', 'pmu_type'],
+            'no (uncore_event_names)'
+        ),
+    MetricName.PLATFORM_DRAM_CLOCKTICKS:
```
ppalucki: Why is it called DRAM_CLOCKTICS ? Is it there any reference that it is just DRAM and not NVM ? Maybe it would be safe to call it IMC_CLOCKTICKS ?

mmucek95: Fixed

PR #405, opened 2020-02-07
Workload mysql-hammerdb stabilize

Damenus:
Mysql-hammerdb workload has often restarted. This PR give more resources to containers and fix bugs, that they do not restart


examples/kubernetes/workloads/mysql-hammerdb/base/mysql-hammerdb.yaml
```diff
@@ -50,20 +50,29 @@ spec:
...
+            valueFrom:
+              resourceFieldRef:
+                containerName: mysql
+                resource: requests.memory
         command:
           - bash
           - -c
           - >
-            local_host='127.0.0.1'
+            # Wait to initialize mysql db
```
mmucek95: Why remove local_host variable?

Damenus: I remove variable and use ip in command below

PR #424, opened 2020-02-28
Configurable uncore metrics

mmucek95:


wca/runners/measurement.py
```diff
@@ -178,6 +178,7 @@ def __init__(
             perf_aggregate_cpus: bool = True,
             enable_derived_metrics: bool = False,
             uncore_event_names: List[str] = [],
```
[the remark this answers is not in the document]
mmucek95: It should :)

wca/runners/measurement.py
```diff
@@ -348,54 +348,137 @@ def _initialize(self) -> Optional[int]:
...
+                                         'missing value to the config file.')
+
+        event_name = configuration[0]
+        assert len(event_name) > 0, 'Uncore event name must not be empty!'
+        event_type = configuration[1]
+        for key_and_value in configuration[2].split(','):
+            separate = key_and_value.split('=')
+            key = separate[0]
+            value = separate[1]
+            if key == 'event':
```
ppalucki: If we want full flexibility we need to allow to provide "config" value - it conflicts with event and umask - so you can either provide (event/umask or config) if both raise an exception.

Second - you need to add, very short description that we only support this forth representation and "mixed representation of event and umask" based on this:

```

OPTIONS
       <command>...
           Any command you can specify in a shell.

       record
           See STAT RECORD.

       report
           See STAT REPORT.

       -e, --event=
           Select the PMU event. Selection can be:

           ·   a symbolic event name (use perf list to list all events)

           ·   a raw PMU event (eventsel+umask) in the form of rNNN where NNN is a hexadecimal event descriptor.

           ·   a symbolically formed event like pmu/param1=0x3,param2/ where param1 and param2 are defined as formats for the PMU in /sys/bus/event_source/devices/<pmu>/format/*

           ·   a symbolically formed event like pmu/config=M,config1=N,config2=K/ where M, N, K are numbers (in decimal, hex, octal format). Acceptable values for each of config, config1 and config2 parameters are defined by
               corresponding entries in /sys/bus/event_source/devices/<pmu>/format/*

                   Note that the last two syntaxes support prefix and glob matching in
                   the PMU name to simplify creation of events accross multiple instances
                   of the same type of PMU in large systems (e.g. memory controller PMUs).
                   Multiple PMU instances are typical for uncore PMUs, so the prefix
                   'uncore_' is also ignored when performing this match.

```

ppalucki: please make that explict in documentaiton/example (e.g. metrics_sources.rst) that yo're only allowing hexadecimal form of providing config/config1/event or mask
(becase 32 is quite different fro 0x32)

ppalucki: as an additional help, you can mention in docs, that user can use: 
```
==> /sys/devices/uncore_imc_0/format/edge <==
config:18

==> /sys/devices/uncore_imc_0/format/event <==
config:0-7

==> /sys/devices/uncore_imc_0/format/inv <==
config:23

==> /sys/devices/uncore_imc_0/format/thresh <==
config:24-31

==> /sys/devices/uncore_imc_0/format/umask <==
config:8-15
```

or this command for cha how to encode paramters into config and config1
```
 head  /sys/devices/uncore_cha_0/format/* 
```

PR #430, opened 2020-03-18
Dram hit ratio algorithm

mmucek95:


wca/scheduler/algorithms/dram_hit_ratio_provision.py
```diff
@@ -0,0 +1,54 @@
...
+
+from wca.metrics import Metric, MetricType
+from wca.scheduler.algorithms.base import BaseAlgorithm, QueryDataProviderInfo
+from wca.scheduler.metrics import MetricName
+from wca.scheduler.types import NodeName
+
+log = logging.getLogger(__name__)
+
+
+THRESHOLD = 0.99
```
Creatone: The threshold could be configurable. ```0.99``` can be a default value for this parameter.

wca/scheduler/algorithms/dram_hit_ratio_provision.py
```diff
@@ -0,0 +1,54 @@
...
+
+
+class DramHitRatioProvision(BaseAlgorithm):
+    def app_fit_node(self, node_name: NodeName, app_name: str,
```
Creatone: There are no tests for this method. 
It's the major feature of your PR, so in my opinion, this should be tested properly.

wca/scheduler/algorithms/dram_hit_ratio_provision.py
```diff
@@ -0,0 +1,54 @@
...
+from wca.scheduler.metrics import MetricName
+from wca.scheduler.types import NodeName
+
+log = logging.getLogger(__name__)
+
+
+THRESHOLD = 0.99
+
+
+class DramHitRatioProvision(BaseAlgorithm):
```
Creatone: There can be a little comment what this algorithm does.

wca/scheduler/algorithms/dram_hit_ratio_provision.py
mmucek95: Done

wca/scheduler/algorithms/dram_hit_ratio_provision.py
mmucek95: Done

wca/scheduler/algorithms/dram_hit_ratio_provision.py
mmucek95: Done

PR #432, opened 2020-03-24
Example experiment runner

mmucek95:


felidadae:
Moved to https://github.com/intel/workload-collocation-agent/pull/437 with kept commits, commits authors from this PR.

PR #426, opened 2020-03-08
Rescheduling in wca-scheduler. Score algorithm.

ppalucki:

PR #435, opened 2020-03-27
Prometheus rules for score2 [wait for #426 to be merged]

mmucek95:


felidadae:
Why the dashboard does not show score2? I checked on graphana internal intel servers.

mmucek95:
> Why the dashboard does not show score2? I checked on graphana internal intel servers.

Please elaborate

PR #436, opened 2020-04-01
Dram hit integration with score

mmucek95:


wca/scheduler/algorithms/score.py
```diff
@@ -62,6 +63,16 @@ def __init__(self, data_provider: ScoreDataProvider,
         super().__init__(data_provider, dimensions, max_node_score, alias)
```
felidadae, line 63: must be divided into base classes, as different parameters

mmucek95: Done

PR #438, opened 2020-04-06
More than four uncore metrics per pmu type

mmucek95:


wca/runners/measurement.py
```diff
@@ -446,44 +446,45 @@ def _prepare_events(self, uncore_events):
...
-        )
-
-        # Wrap with derived..
-        if enable_derived_metrics:
-            self._uncore_derived_metrics = UncoreDerivedMetricsGenerator(
-                self._uncore_pmu.get_measurements)
-            self._uncore_get_measurements = self._uncore_derived_metrics.get_measurements
-        else:
-            self._uncore_get_measurements = self._uncore_pmu.get_measurements
+        for event_groups in uncore_events:
```
ppalucki, line 453: we can trick here to be backward compatible - 
if first of elements is a string, then 

uncore_events = [uncover_events]

before the for

wca/runners/measurement.py
```diff
@@ -177,7 +177,7 @@ def __init__(
             event_names: List[str] = [],
             perf_aggregate_cpus: bool = True,
             enable_derived_metrics: bool = False,
-            uncore_event_names: List[str] = [],
+            uncore_event_names: List[List[str]] = [],
```
ppalucki, line 180: if we use the trick to allow single list, the type has to be:
```python
uncore_event_names: List[Union[List[str], str]]
```

ppalucki: this will allow to configure wca:
```yaml
uncore_event_names:
- x1 
- x2
```

or 
```yaml

uncore_event_names:
- [x1, x2]
```
or
```yaml 
uncore_event_names:
- [x1, x2]
- [x3, x4]
```

PR #441, opened 2020-04-08
StaticNode can now discovery Docker containers as tasks with "directory" option (without metadata) + fix for DDR3 (only DDR4 was handled) 

ppalucki:
This is not full support for Docker containers, there is no metadata from (no real name and labels) - only container ids.

Check "configs/extra/static_measurements_directory.yaml" as an example

or run it like this:

```sh
 sudo env PYTHONPATH=. `which python` wca/main.py -c $PWD/configs/extra/static_measurements_directory.yaml -0
```


wca/platforms.py
```diff
@@ -262,6 +263,7 @@ def get_platform_static_information(strict_mode: bool) -> Measurements:
             dmidecode_str = dmidecode_raw.decode("utf-8")
             nvm_dimm_count, ram_dimm_count, nvm_dimm_size, \
                 ram_dimm_size, ram_dimm_speed = _parse_dmidecode_output(dmidecode_str)
+            print('debug:', nvm_dimm_count, ram_dimm_speed, nvm_dimm_size)
```
mmucek95, line 266: Please don't use print here.

ppalucki: thx !!!

PR #443, opened 2020-04-09
External metrics (based on previous PR)

ppalucki:
Allow to collect external metrics providing sub process command to collect and regexp to parse output from external process.

Documentation and idea described here: https://github.com/ppalucki/owca/blob/ppalucki/external-metrics/docs/api.rst#external


ppalucki:
Suspended for now - not required anymore, was only used for testing purposes.

PR #448, opened 2020-04-20
Limiting cpu for experiments

mmucek95:


Damenus:
Please, you should add in config example using this new feature

examples/kubernetes/wca-scheduler/config.yaml
```diff
@@ -14,6 +14,7 @@ algorithm: !Score
     - membw_write
     - membw_read
     - wss
+  cpu_scale_factor: 1
```
Damenus, line 17: Here, it should be a float, no int.

PR #453, opened 2020-04-27
Dockerfiles improvements

mmucek95:


Makefile
```diff
@@ -201,3 +205,18 @@ tester:
 generate_docs:
 	@echo Generate documentation.
 	$(call execute_in_venv, env PYTHONPATH=. python util/docs.py)
+
+hadolint_check:
```
felidadae, line 209: if wants to check all Dockerfiles you could use find tool (find . -name Dockerfile)

examples/workloads/stress_ng/Dockerfile
```diff
@@ -20,8 +20,8 @@ RUN yum install -y gcc git make wget patch
 RUN wget https://github.com/ColinIanKing/stress-ng/archive/V0.10.08.tar.gz
 RUN tar -xzf V0.10.08.tar.gz
 WORKDIR /stress-ng-0.10.08
-ADD APM.patch APM.patch
-RUN cat APM.patch | patch -p 1
+COPY APM.patch APM.patch
```
Damenus, line 23: Why COPY is better then ADD?

Dockerfile
```diff
@@ -34,11 +34,11 @@ RUN yum -y update && yum -y install python36 python-pip which make git wget
 # The official way to build impctl is used below,
 # based on instruction from https://github.com/intel/ipmctl#build,
 # where ndctl, libsafec is dependency of ipmctl
-RUN (cd /etc/yum.repos.d/; \
-        wget https://copr.fedorainfracloud.org/coprs/jhli/ipmctl/repo/epel-7/jhli-ipmctl-epel-7.repo; \
-        wget https://copr.fedorainfracloud.org/coprs/jhli/safeclib/repo/epel-7/jhli-safeclib-epel-7.repo)
+WORKDIR /etc/yum.repos.d
+RUN wget https://copr.fedorainfracloud.org/coprs/jhli/ipmctl/repo/epel-7/jhli-ipmctl-epel-7.repo
```
Damenus, line 38: Can be there one RUN with the two wget?

examples/workloads/stress_ng/Dockerfile
mmucek95: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#add-or-copy

Dockerfile
mmucek95: Yes, there can be.

PR #455, opened 2020-05-11
Admission controller - annotating pods

mmucek95:


examples/kubernetes/monitoring/prometheus/prometheus_rule.app.yaml
```diff
@@ -63,6 +63,10 @@ spec:
         - record: app_contention_sensitivity
           expr: clamp_max(app_memory_intensity * app_memory_utilization / app_rw, 5)  # for DEMO1
 
+        # Requested to WSS ratio in %
+        - record: app_pmem_ratio
```
felidadae, line 67: why that name? should not be app_wss_to_mem_ratio ?

requirements.txt
```diff
@@ -9,5 +9,6 @@ pytest==5.3.5                    # 2020.02.07
 pytest-cov==2.8.1                # 2019.10.05
 flake8==3.7.9                    # 2019.10.28
 setuptools_scm==3.4.3            # 2020.02.07
-flask==1.1.1		         # 2019.12.15
-gunicorn==20.0.4		 # 2019.12.18
+flask==1.1.1		             # 2019.12.15
```
felidadae, line 12: wrong indentation

wca/admission_controller/wss_data_provider.py
```diff
@@ -0,0 +1,41 @@
...
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License
+
+from wca.prometheus import Prometheus
+
+WSS_QUERY = 'max(max_over_time(task_wss_referenced_bytes[3h])) by (app)'
```
felidadae, line 17: can be configured in config file? (this and two below lines)
i think in scheduler there was option to do so, i think it would be better to stick to the previous way of doing things.

PR #456, opened 2020-05-22
Configs, prom rules

felidadae:
- score rules improved:
  - (ppalucki scaled version for R/W)
  - add membw_flat
  - fix naming to stick with presentation (20.05.2020)
- kustomize fix for redis
  - lower latency
- Change wca-scheduler config to use metrics created in prom rules (previously we had two sources of truth)
- as now score is not negative (just remove mutlipling by (-1)), I adjust proper line in Score algorithm
- add to prom rules safety capacities of PMEM
- finish improving config for wca-scheduler
- remove wrong defaults for Queries in DataProvider


examples/kubernetes/wca-scheduler/config.yaml
```diff
@@ -6,31 +6,31 @@ kubeapi: &kubeapi !Kubeapi
...
+  score_target: 1.5
   max_node_score: 10000.
   dimensions:
     - mem
     - cpu
     - membw_write
     - membw_read
     - wss
-  cpu_scale_factor: 1.0
+  cpu_scale_factor: 0.8
```
Damenus, line 17: We shouldn't unify setting max utilization for resources? CPU has `cpu_scale_factor` param, but MBW has Queries multiplication by number.

examples/kubernetes/monitoring/prometheus/prometheus_rule.score.yaml
```diff
@@ -119,14 +119,14 @@ spec:
       expr: count(task_up) by (app)
 
     # very low metric
-    - record: task_memory_rw_ratio
-      expr: rate(task_offcore_requests_demand_data_rd[10m]) / (rate(task_offcore_requests_demand_data_rd[10m]) + rate(task_offcore_requests_demand_rfo[10m]))
+    - record: app_memory_rw_ratio
+      expr: (rate(task_offcore_requests_demand_data_rd[10m]) + rate(task_offcore_requests_demand_rfo[10m])) / (rate(task_offcore_requests_demand_data_rd[10m]) + 2*rate(task_offcore_requests_demand_rfo[10m]))
```
Damenus, line 123: Why you multiply by 2?

felidadae: its ppalucki equation

examples/kubernetes/wca-scheduler/config.yaml
felidadae: i think the same

examples/kubernetes/monitoring/prometheus/prometheus_rule.score.yaml
Damenus: Ok

PR #461, opened 2020-06-03
Working set size - stabilization

mmucek95:


wca/wss.py
```diff
@@ -22,41 +23,127 @@
...
+    def _discover_pids(self):
         pids = set(self.get_pids(include_threads=False))
         all_pids = os.listdir('/proc')
-        all_pids = set(filter(lambda x: x.isdecimal(), all_pids))
+        all_pids = set(map(str, filter(lambda x: x.isdecimal(), all_pids)))
         pids = pids & all_pids
-        log.debug('calculating wss for pids %s', pids)
+        return pids
+
+    def _check_if_stable(self, referenced, resgroup, mongroup_name):
```
ppalucki, line 54: I've not added description in my PoC, but can you do this.
Remember a comment should express intention not mechanic of the function. 

And what I see it is responsible for update internal ``stable`` counter. Please consider renaming this function to something like "_update_stable_counter"

wca/wss.py
```diff
@@ -22,41 +23,127 @@
...
+        for socket_dir in socket_dirs:
+            file = resgroup.get_event_file(socket_dir, 'mbm_total_bytes', mongroup_name)
+            current[socket_dir] = float(open(file).read())
+
+        if self.prev_value:
+            value = {s: (current[s] - self.prev_value[s]) / self.interval_s / 1000000
+                     for s in self.prev_value.keys()}
+            self.bw_values.append(value)
+        self.prev_value = current
+        bw_mbs = self.get_total_bw()
```
ppalucki, line 66: You could use existing mechanism of collecting MBW from resctrl.py, couldn't you?
I'm ok with existing because it still a PoC but just want to know was is too hard or impossible?

wca/wss.py
```diff
@@ -22,41 +23,127 @@
...
 
-        if self._cycle % self.reset_interval == 0:
-            for pid in pids:
-                try:
-                    with open('/proc/{}/clear_refs'.format(pid), 'w') as f:
-                        f.write('1\n')
-                except FileNotFoundError:
-                    log.warning('pid does not exist for clearing refs - ignoring!')
-                    pass
+        referenced = int(referenced/1024)  # scale as megabytes
```
ppalucki, line 129: There is a bug here, first you're dividing by 1000 to multiple by 1024 in next line. Very suspicious.

wca/wss.py
```diff
@@ -22,41 +23,127 @@
...
+        self._check_if_stable(referenced, resgroup, mongroup_name)
+
+        if self.stable_cycles != 0 and self.stable == self.stable_cycles:
+            self.stable = 0
+            should_reset = True
+            measurements[MetricName.TASK_WORKING_SET_SIZE_BYTES] = referenced
```
ppalucki, line 135: I think, that if we do not have, stable result then we should return last stable value (if any exists) - your way, we're risking that the values between will be extrapolated by Prometheus.
Maybe we should add another metric to be able to conclude how stable the task_working_set_size_bytes is.

mmucek95: Added the last stable value if a cycle is not stable.

wca/wss.py
mmucek95: Fixed.

wca/runners/measurement.py
```diff
@@ -191,9 +191,9 @@ def __init__(
             task_label_generators: Optional[Dict[str, TaskLabelGenerator]] = None,
             allocation_configuration: Optional[AllocationConfiguration] = None,
             wss_reset_interval: int = 0,
+            wss_stable_duration: int = 30,
```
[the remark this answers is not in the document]
mmucek95: Fixed.

PR #462, opened 2020-06-03
README for score algorithm

felidadae:


docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
+************************
+Score algorithm overview
+************************
+
+We constructed a heuristic for automatic assessment of how well a workload match a node with
+Intel PMEM memory installed ran in 2LM mode. The heuristic is trying to reach two goals:
+1) use as much of memory as possible on PMEM nodes 2) minimize a chance of worsening workloads performance
```
ppalucki, line 7: add line breaks

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+(comparing to DRAM). The 2) is approached with idea that performance can be degraded mostly when any of: memory bandwidth,
```
ppalucki, line 8: Add another line break and I suggest with replace '2)' with 'performance impact is reduced by...'.

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+L4 cache (in 2LM mode DRAM runs as a next level cache) shared resources are saturated on a node.
+
+*The heuristic assigns to each workload a single positive rational number*, thus creating a sorted list, if
```
ppalucki, line 11: I do not understand "if taking into consideration all workloads".

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+taking into consideration all workloads. That single number, we call a workload's **score**.
+**Lower** the number, **better** a workload fits the PMEM node according to our heuristic.
+The value of score does not depends on different workloads, but only on workload innate features and
```
ppalucki, line 14: I do not understand "does not depends on differnt workloads" - let's repalce it with:
``The value of score depends on workload innate features and resources capacity of target average PMEM node``.

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+
+We prepared Grafana dashboard `graphana dashboard <../examples/kubernetes/monitoring/grafana/2lm_dashboards/2lm_score_dashboard.yaml>`_
+for visualization of the results mentioned in `Scores for our testing workloads`_.
+
+Limitations
+###########
+
+There are few limitations of our solution, which depending on usage can constitute a problem:
+
+- no support for **versions** of statefulset/deployment,
```
ppalucki, line 229: append this: " ... which means, that statefulset or deployment is considered as the same workload as long as the name of statefulset/deployment matches."""

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+- due to used mentioned in `Configuring the Score` prometheus functions, requirements of some
```
ppalucki, line 230: remove "due to used mentioned in `Configuring the Score` prometheus functions, "

docs/score_algorithm.rst
```diff
@@ -0,0 +1,233 @@
...
+workload can be overestimated, e.g. if workload is wrongly configured and keeps restarting after a short period of time
+- as we take CPU/MEM requirements from Kubernetes defined limits, if workload does not have that data defined,
```
ppalucki, line 232: reverse the limitation, we only supports with defined cpu/memory

PR #465, opened 2020-06-08
Improvements for running runner_analyzer scripts only on nodes node10*

felidadae:
Must be rebased on master


ppalucki:
Conflicts with latest version of runner.

PR #483, opened 2020-07-10
Deploy cadvisor alongside wca

wacuuu:


ppalucki:
One more thing, please move Dockerfile.cadvisor to examples/kubernetes/monitoring/cadvisor is possible.

PR #496, opened 2020-08-20
vmstat(s) global and per NUMA node for all keys

ppalucki:
Only limited number of vmstats were available currently.

This PR collects all keys from both:
- /proc/vmstat
- /sys/devices/system/node/nodeX/vmstat

and exposes them as new two metrics

platform_vmstat with key label
and 
**platform_node_vmstat** with numa_node and key labels


Damenus (approved):
LGTM

PR #502, opened 2020-09-08
AnnotationService improvements

felidadae:
Adding also WCA config file, as wss_reset_interval set to -1 is needed.


mmucek95:
Moved to #547

PR #505, opened 2020-09-10
Hmem analyzer

Damenus:


felidadae (changes requested):
To fix:
Scenario.scenario_type -> Scenario.experiments_type
to be a little more consistent in naming (or rename ExperimentType to ScenarioType).

Scenario.workloads is not needed and even a little bit misleading, just use workloads_count, add in the comment example [{workload_a: 1, workload_b: 3}]

Generated pdf:
- it would be great to add experiment type description to the pdf.

felidadae (approved):

PR #513, opened 2020-09-29
WSS improvements (optional fraction thresholds and small fix)

ppalucki:
- Thresholds now called mbw_fraction and ref_fraction are independent and optional.

Additionally:

- small fix for WSS, just record first value after on of the threshold is met and better logs

Again fix warning about non consecutive NUMA nodes (it was false positive).

New features:

I. ability to run in four modes:

1. referenced_bytes only read only (wss_reset_cycles=0 and wss_stable_cycles=0) 
2. referenced_bytes and working set size metrics with cyclic reset (wss_reset_cycles>0 and wss_stable_cycles<0) **recommended** - least intrusive 
3. referenced_bytes and working set size metrics with reset after stable period (wss_reset_cycles>0 and wss_stable_cycles<0 - negative number here suggests not reseting after stable cycle ends)
  
II. referenced bytes (kernel time) and access bit clear process overhead (new metrics about time spent for this)


felidadae (changes requested):
wss_reset_interval -> wss_reset_cycles
wss_stable_duration -> wss_stable_cycles

Why keeping old wrong logic `lower_than_ref`? It cannot even recognize linear pattern (or maybe on purpose?).

_update_stable_counter is confusing, i mean not in theory, i read code and and it was just confusing.
Could you rename to _check_stability or sth like that?

Is that ok, that referenced is counted in the whole window (of width `interval`=cycle_duration) and membw is calculated as metric per second? dont we want to keep the same window for both? 
also names delta_referenced and delta_membw are a little bit confusing as there are not simple deltas, but also normalized per second --> (curr-prev)/interval.

ppalucki:
@felidadae 

0. naming: *_cycles/update_stable_counter/delta_referenced/delta_membw I will fix
1. on old "logic" is not wrong, but is different - I wanted to have that in backup in case, that RDT is not available - but you're right it is better to remove to prevent confusion
2. of course they should be,, you're right, but I thought both are per second, I missed that - *nice catch*, gonna fix

felidadae:
i think it would be less confusing:
delta_referenced = curr_referenced - prev_referenced
then in the checking the flag
delta_referenced < curr_membw * **interval** * wss_mbw_fraction

curr_membw_delta is not used, can be removed

felidadae (approved):

PR #517, opened 2020-10-02
Update rules with labels.

felidadae:
Similar to https://github.com/intel/workload-collocation-agent/pull/512
to make it possible to use both wca/cadvisor rules in the same time on a single cluster.

### DIFF
--- a/examples/kubernetes/monitoring/prometheus/prometheus_rule.score.yaml
+++ b/examples/kubernetes/monitoring/prometheus/prometheus_rule.score.yaml
@@ -17,49 +17,59 @@ spec:
       expr: 'sum(platform_topology_cpus) by (node)'
       labels:
         dim: cpu
+        source: wca
     # mem
     - record: node_capacity_unsafe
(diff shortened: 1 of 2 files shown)


examples/kubernetes/monitoring/prometheus/prometheus_rule.score.yaml
```diff
@@ -121,13 +135,21 @@ spec:
       expr: 'task_mbw_read + task_mbw_write'
 
     - record: app_cpu
-      expr: 'max(max_over_time(task_requested_cpus[7d])) by (app)'
+      expr: 'max(max_over_time(task_requested_cpus[2m])) by (app)'
```
wacuuu, line 138: I guess changed range was for testing only or this should be so short?

felidadae: fixed

wacuuu (approved):

PR #534, opened 2020-10-16
Fix sched debug stats - collect per thread basis

ppalucki:
Current method only measures sched stats of main thread (starting threads but ignores all children threads).


Damenus (approved):
LGTM

PR #547, opened 2020-11-03
AnnotationService improvements

mmucek95:


ppalucki (approved):

wca/admission_controller/app_data_provider.py
```diff
@@ -18,7 +18,9 @@
 
 @dataclass
 class Queries:
-    APP_WSS_TO_MEM_RATIO = 'max(max_over_time(app_wss_to_mem_ratio[3h])) by (app)'
+    APP_WSS: str = \
+        'app_req{dim="wss"}'
+    APP_RSS: str = 'app_rss'
```
ppalucki, line 23: maybe not used directly but it should be ```app_req{dim="mem"}```

mmucek95: Done

PR #553, opened 2020-11-17
Join scheduler and annotator

mmucek95:


tests/scheduler/test_server.py
```diff
@@ -1,59 +1,67 @@
...
-        output = annotating_service._mutate_pod()
-        assert output.json == {
-                "response": {
-                    "allowed": True,
-                    "status": {"message": 'Patching pod'},
-                    "patch": base64.b64encode(str(patch).encode()).decode(),
-                    "patchtype": "JSONPatch",
-                }
-            }
+<<<<<<< HEAD:tests/scheduler/test_server.py
```
Creatone, line 1: Unnecessary

wca/scheduler/server.py
```diff
@@ -53,10 +63,84 @@ def reschedule_interval(self, interval: Numeric(0, 60)):
...
+                    "allowed": True,
+                    "status": {"message": message},
+                    "patch": base64.b64encode(str(patch).encode()).decode(),
+                    "patchtype": "JSONPatch",
+                }
+            }
+        )
+
+    def _get_memory_type(self, wss: str, rss: str):
+        ratio = float(wss) / float(rss) * 100
```
Creatone, line 81: Possible `ZeroDivisionError`

Creatone (changes requested):

tests/scheduler/test_server.py
mmucek95: Removed

wca/scheduler/server.py
mmucek95: Fixed

Creatone (approved):

